PR #26, opened 2020-09-18
add returnwordlist function

### LINKED ISSUE #23: Create a TypeScript function that returns a wordlist given an Excel file
In order to consider this "done", there should be:

 - [ ] a TypeScript function in `src/worker/read-wordlist.ts`
 - [ ] a passing test case that calls this function with an Excel file

|word|count|
|-|-|
|TŦE|13644|
|E|9134|
|SEN|4816|

The wordlist should be a data structure that looks like this:

```json
[
  ["TŦE", 13644],
  ["E", 9134],
  ["SEN", 4816]
]
```

xqiu1:
add an incomplete returnWordlist function

Fixes #23


src/worker/read-wordlist.ts
```diff
@@ -0,0 +1,12 @@
+import readXlsxFile from "read-excel-file";
+
+export function returnWordlist(
+  ExcelFile: Buffer
+): Promise<Array<[string, number]>> {
```
eddieantonio:
```suggestion
): Promise<[string, number][]> {
```

src/worker/read-wordlist.ts
```diff
@@ -0,0 +1,12 @@
...
+  let wordlist: Array<[string, number]> = [];
```
eddieantonio:
```suggestion
  let wordlist: [string, number][] = [];
```

eddieantonio (approved):
LGTM! But use the `T[]` notation instead of the `Array<T>` notation in the future!

PR #28, opened 2020-09-19
feature: invoke keyman compiler

### LINKED ISSUE #19: Run the compiler within the Web Worker
Simply invoke the Keyman Lexical model compiler and produce JavaScript output.

eddieantonio:
Invokes the Keyman compiler from the tests, but should invoke it from the worker soon 👍 

Resolves #19


test/test-generate-wordlist.ts
```diff
@@ -0,0 +1,32 @@
...
+        ["TŦE", 13644],
+        ["E", 9134],
+        ["SEN", 4816],
+      ]),
+    ],
+  });
+  // it will compile to a JavaScript function!
+  t.regex(code, /\bfunction\b/);
+
+  t.notThrows(function () {
```
IanSeng: **suggestion** can we use arrow function ?

test/test-generate-wordlist.ts
```diff
@@ -0,0 +1,32 @@
...
+    // We should be able to eval that function and get its code:
+    new Function(code);
```
IanSeng, line 31: same as above

IanSeng (changes requested):

test/test-generate-wordlist.ts
eddieantonio: ...not for this one. The purpose of this is to **parse the JavaScript code** generated by the lexical model compiler. See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/Function

test/test-generate-wordlist.ts
eddieantonio: Yes for this one! ...but

IanSeng (approved):
👍

PR #42, opened 2020-09-19
[#10]feat: Setup precommit hooks with Husky

### LINKED ISSUE #10: Setup precommit hooks with Husky

IanSeng:
# Summary 
- implemented husky pre-commit hooks 
- updated scripts in package.json

Closes #10

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
     "eslint": "^7.7.0",
     "eslint-config-prettier": "^6.11.0",
     "eslint-plugin-prettier": "^3.1.4",
+    "husky": "^4.3.0",
     "prettier": "^2.1.1",
     "rollup": "^2.26.5",
     "rollup-plugin-livereload": "^1.3.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -42,6 +43,7 @@
   "dependencies": {
     "@rollup/plugin-typescript": "^5.0.2",
     "@tsconfig/svelte": "^1.0.10",
+    "husky": "^4.3.0",
```
eddieantonio: This should be `devDependency`.

Honestly, in our case it's more a matter of principle than something that is technically wrong, so I won't block the PR.

eddieantonio (approved):
LGTM!

But, it would be _extra nice_ if you mentioned husky in `.github/CONTRIBUTING.md`!

package.json
IanSeng: I did --save-dev but no clue why it's here but nice catch, I will fix it

Issue #18, opened 2020-09-18
Create Figma wireframe(s)


eddieantonio:
![Landing page](https://user-images.githubusercontent.com/2294397/93635953-54c55400-f9b0-11ea-9df7-4a32ebdbea5b.png)
![Basic info page](https://user-images.githubusercontent.com/2294397/93635950-53942700-f9b0-11ea-9f01-9a104314a2e5.png)
![Dictionary sources page](https://user-images.githubusercontent.com/2294397/93635948-5262fa00-f9b0-11ea-94e9-43e6222f550f.png)

pranavnarang:
Link to Figma project: https://www.figma.com/file/0ifTqGA2I3Y216eCJxh1rr/Keyman-Predictive-Text?node-id=0%3A1

PR #50, opened 2020-09-20
change: clean up global styles

### LINKED ISSUE #44: Start CSS with Keyman brand colours
These are the same colours as in the presentation!

eddieantonio:
This cleans up some of the Keyman brand colours introduced by @jienius. It also uses the font stack that @pranavnarang used in the Figma mockups (#18). I would appreciate a review from you both!

resolves #44

### DIFF
--- a/src/app/App.svelte
+++ b/src/app/App.svelte
@@ -7,18 +7,32 @@
 
 <style>
   :global(:root) {
-    --blue: #6fbbd4;
-    --blue-light: #b3dbe7;
-    --red: #b90529;
-    --red-light: #de929c;
-    --orange: #f18407;
-    --orange-light: #fbc88f;
+    /* Keyman brand colours, derived from https://keyman.com/ */
+    --blue: #6FBBD4;
+    --blue-light: #B3DBE7;
+    --red: #B90529;
+    --red-light: #DE929C;
+    --orange: #F18407;
+    --orange-light: #FBC88F;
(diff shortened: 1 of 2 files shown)


jienius (approved):
The Roboto font makes it looks much nicer now!

pranavnarang (approved):
LGTM!

src/app/App.svelte
```diff
@@ -7,18 +7,32 @@
...
+    font-family: var(--main-font), sans-serif;
+  }
+
+  :global(h1, h2, h3, h4, h5, h6) {
+    font-family: var(--main-font), sans-serif;
+  }
+
+  :global(pre, code, samp) {
+    font-family: var(--mono-font), monospace;
   }
```
IanSeng, line 36: **Question**: Do we have global stying for body text? I would strongly recommend implementing a utility class for the purpose of having a clean and tidy looking css. Other than that we could also package our global styling to make this global value file looks cleaner

IanSeng (commented):
Opps, I think I am too late 🙃

src/app/App.svelte
eddieantonio: Yeah, I think I'd prefer a global CSS file rather than tucking the CSS within each component, but I'll create a new issue for that!

If we want to incorporate that with Svelte component styles, we'll need a bit more work...

PR #56, opened 2020-09-20
add generate zip function

### LINKED ISSUE #47: Find a zip file generating library that works in the browser
To close this, it should be added to package.json and works should be invoked in the unit tests. WITHOUT TOUCHING THE FILE SYSTEM (no output paths).

Should take plain text files and generate a zip file.

xqiu1:
Fixes #47


package.json
```diff
@@ -44,6 +44,8 @@
   "dependencies": {
     "@rollup/plugin-typescript": "^5.0.2",
     "@tsconfig/svelte": "^1.0.10",
+    "install": "^0.13.0",
```
eddieantonio: Is there a reason this module got... [ahem]... installed? https://www.npmjs.com/package/install

Or was it added by mistake?

eddieantonio (changes requested):
I think a package got installed by mistake -- please remove it: https://classic.yarnpkg.com/en/docs/cli/remove/

Another thing is, did you forget to add some files?

eddieantonio (changes requested):
The test and the implementation files are still missing! >.<

eddieantonio (approved):
LGTM

src/worker/generate-zip.ts
```diff
@@ -0,0 +1,5 @@
+import JSZip = require("jszip");
```
eddieantonio: My bad: this is the way that the compiler actually wants things to be compiled 😩:


```suggestion
import * as JSZip from "jszip";
```

This will solve the failing build: https://github.com/eddieantonio/predictive-text-studio/pull/56/checks?check_run_id=1140747500#step:3:47

eddieantonio (changes requested):

xqiu1:
Still got an eroor when I run yarn test: 
test/test-generate-wordlist.ts(11,8): error TS2307: Cannot find module '@predictive-text-studio/lexical-model-compiler' or its corresponding type declarations.

eddieantonio (approved):
LGTM!

(we finally figured out the imports! 😅 )

Issue #48, opened 2020-09-19
Find an appropriate router for Svelte


eddieantonio:

pranavnarang:
For whoever picks up this issue, found this helpful: https://dev.to/lukocastillo/svelte-3-how-to-integrate-with-svelte-routing-4j3b.

tldr: most popular svelte routing library at the moment is https://www.npmjs.com/package/svelte-routing

PR #58, opened 2020-09-20
Feat/lang info page

### LINKED ISSUE #29: Make a executable mockup of Language Info page (without tests)
Create a component in Svelte that displays the figma design of the language information page

pranavnarang:
Barebones UI components for the language information page, (will) resolve #29 

Main action items left:
  1. Add JSDoc Notation to functions
  2. Create/refactor button component
  3. Use variable names for colors and fonts
  4. Cypress testing 

Will likley need to wait for #48 as routing needs to be setup before this page can be accessed. Right now it replaces what is on `App.svelte`.


src/app/Components/LanguageInfo.svelte
```diff
@@ -0,0 +1,54 @@
+<script lang="ts">
+  import InputField from "./InputField.svelte";
+  export let properties;
```
pranavnarang: Need to add type here

src/app/App.svelte
```diff
@@ -1,13 +1,14 @@
 <script lang="ts">
-  export let name: string;
-
+  // export let name: string;
```
eddieantonio:
```suggestion
```

src/app/App.svelte
```diff
@@ -1,13 +1,14 @@
...
+  import Languages from "./Languages.svelte";
 </script>
 
 <main role="application">
-  <h1> Welcome to {name}!</h1>
+  <!-- <h1> Welcome to {name}!</h1>
```
eddieantonio:
```suggestion
```

src/app/App.svelte
```diff
@@ -1,13 +1,14 @@
...
 
   <p>Hi, my name is Xuechun!</p>
   <p> Hi my name is Eddie! </p>
   <p>Pranav</p>
-  <h2>Ian</h2>
+  <h2>Ian</h2> -->
```
eddieantonio:
```suggestion
```

src/app/App.svelte
eddieantonio: GitHub wouldn't let me select the code inbetween this and the `-->`, but basically, delete that too!

eddieantonio (commented):
I didn't do a thorough review (as I think that's inappropriate for now).

If there's old, placeholder code getting in your way, please just delete it! 🔥 🔥 🔥  If we miss it, that's what the commit log is for! :D

src/app/Components/Button.svelte
```diff
@@ -0,0 +1,95 @@
...
+    isOutlined && (className += " outlined");
+    hasDropShadow && (className += " shadow");
+  };
+
+  createStyles();
+</script>
+
+<style>
+  button {
+    margin: 0px 10px;
```
IanSeng: **[Suggestion]**: Rem instead of pixel

src/app/Components/Button.svelte
```diff
@@ -0,0 +1,95 @@
...
+    border-radius: 10px;
```
IanSeng: Same as above, and most of the styling bellow

src/app/Components/Button.svelte
```diff
@@ -0,0 +1,95 @@
...
+  .medium {
+    padding: 10px 20px;
+  }
+
+  .large {
+    padding: 20px 40px;
+  }
+
+  .grey {
+    background-color: #bdbdbd;
```
IanSeng: **[Suggestion]**: create a global variable for the reason of reusability and consistency

src/app/Components/Button.svelte
```diff
@@ -0,0 +1,95 @@
...
+  .small {
+    padding: 5px 10px;
+  }
+
+  .medium {
+    padding: 10px 20px;
+  }
+
+  .large {
+    padding: 20px 40px;
```
IanSeng: **[Question]**: I am not sure where we have landed, if we have decided to implement utility classes, we shouldn't have these in local styling.

src/app/Components/Button.svelte
```diff
@@ -0,0 +1,95 @@
...
+    background-color: #f1f1f1;
+  }
+
+  .shadow {
+    box-shadow: 0px 25px 30px rgba(0, 0, 0, 0.15);
+  }
+</style>
+
+{#if text && callbackFunction}
+  <button on:click={callbackFunction} class={className}> {text} </button>
```
IanSeng: **Suggestion**: a test-id should be added into this button if we moving forward with using this as a reusable component

src/app/Components/InputField.svelte
```diff
@@ -0,0 +1,14 @@
+<script lang="ts">
+  export let label: string;
+  export let id: string;
+  export let value: any;
+  export let subtext: string;
+</script>
+
+<div class="input">
+  <h3>{label}</h3>
```
IanSeng: **Suggestion**: if this is the header of the page this h3 is not a good practice for accessibility

src/app/Components/LanguageAvatar.svelte
```diff
@@ -0,0 +1,56 @@
+<script lang="ts">
+  export let name: string;
+  export let selectedLanguage: string = "";
+
+  export let isSelected: boolean = name === selectedLanguage;
```
IanSeng: **[Question]**: this line of code seems to be very confusing?

src/app/Components/LanguageAvatar.svelte
```diff
@@ -0,0 +1,56 @@
...
+  export let className: string = "avatar";
+  if (isSelected) {
+    className = className + " selected";
```
IanSeng: **[Question]**: I would avoid magic string as much as possible

src/app/Components/LanguageAvatar.svelte
```diff
@@ -0,0 +1,56 @@
...
+  .avatar {
+    width: 50px;
+    height: 50px;
+    display: flex;
+    background-color: #fff;
+    align-items: center;
+    justify-content: center;
+    border-radius: 100%;
+    margin-top: 40px;
+    font-family: Cabin, sans-serif;
```
IanSeng: **[Suggestion]**: font-family has appeared quite a few times, can we use it in global stying

src/app/Components/LanguageAvatar.svelte
```diff
@@ -0,0 +1,56 @@
...
+    console.log(languageName + " button selected");
+  };
+</script>
+
+<style>
+  .avatar {
+    width: 50px;
+    height: 50px;
+    display: flex;
+    background-color: #fff;
```
IanSeng: Same as above, use a variable

src/app/Components/LanguageAvatar.svelte
```diff
@@ -0,0 +1,56 @@
...
+   * @return {void}
+   */
+  const handleClick = (languageName: string): void => {
+    console.log(languageName + " button selected");
+  };
+</script>
+
+<style>
+  .avatar {
+    width: 50px;
```
IanSeng: Ram instead of pixel

src/app/Components/LanguageAvatar.svelte
```diff
@@ -0,0 +1,56 @@
...
+    justify-content: center;
+    border-radius: 100%;
+    margin-top: 40px;
+    font-family: Cabin, sans-serif;
+    font-size: 20px;
+  }
+
+  .avatar:hover {
+    cursor: pointer;
+    box-shadow: 0px 10px 20px rgba(0, 0, 0, 0.1);
```
IanSeng: **[Suggestion]**: A hover styling effect variable can be done to reduce code duplication

src/app/Components/LanguageInfo.svelte
```diff
@@ -0,0 +1,58 @@
+<script lang="ts">
+  import InputField from "./InputField.svelte";
+  export let properties: any;
+</script>
+
+<style>
+  .language__info {
```
IanSeng: **[Question]**: I love the BEM naming convention, but if we are using CSS these doesn't make a different

src/app/Components/LanguageSources.svelte
```diff
@@ -0,0 +1,60 @@
...
+
+  th {
+    text-align: left;
+    background-color: #b6b6b6;
+    padding: 20px 20px 20px 5px;
+  }
+
+  table {
+    border-collapse: collapse;
+    width: 1400px;
```
IanSeng: Same as above, use rem

src/app/Components/LanguageSources.svelte
```diff
@@ -0,0 +1,60 @@
...
+    flex-direction: row;
+  }
+
+  .table__row--actions p {
+    margin-left: 10px;
+  }
+</style>
+
+<div class="language__sources">
+  <table>
```
IanSeng: **[Suggestion]**: this could be a list rather than a table, on top of that table is relatively harder for adding accessibility functions

src/app/Components/Sidebar.svelte
```diff
@@ -0,0 +1,23 @@
...
+  export let languages: object[];
+  export let selectedLanguage: string;
+
+  import LanguageAvatar from "./LanguageAvatar.svelte";
+</script>
+
+<style>
+  .sidebar {
+    height: 100%;
+    background-color: #454545;
```
IanSeng: same as above, use a variable

src/app/Components/InputField.svelte
```diff
@@ -0,0 +1,14 @@
+<script lang="ts">
+  export let label: string;
```
IanSeng: **[Suggestion]**: If we are using this variable locally, export can be removed.

src/app/Components/LanguageAvatar.svelte
```diff
@@ -0,0 +1,56 @@
...
+  .selected {
+    background-color: #0099ff;
+    color: #fff;
+    box-shadow: 0px 10px 20px rgba(0, 0, 0, 0.1);
+  }
+</style>
+
+<div
+  class={className}
+  on:click={() => {
```
IanSeng: **[Suggestion]**: if im not wrong, this extra bracket can be removed

src/app/Components/LanguageInfo.svelte
```diff
@@ -0,0 +1,58 @@
+<script lang="ts">
+  import InputField from "./InputField.svelte";
+  export let properties: any;
```
IanSeng: **[Suggestion]**: type any should be avoided unless this is a throw away code

src/app/Components/LanguageInfo.svelte
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  .language__info-right {
+    margin-left: 300px;
+    display: flex;
+    flex-direction: column;
+    order: 2;
+  }
+
+  img {
```
IanSeng: **[Suggestion]**: styling with anchor tag is will be a troublesome when it comes to reusing this component

src/app/Languages.svelte
```diff
@@ -0,0 +1,212 @@
...
+  .languages__container--actions {
+    display: flex;
+    flex-direction: row;
+  }
+
+  .languages__container--content {
+    margin-top: 25px;
+  }
+
+  button {
```
IanSeng: same as above

src/app/Languages.svelte
```diff
@@ -0,0 +1,212 @@
...
+      <div class="languages__container--header">
+        <img
+          src="https://keyman.com/cdn/deploy/img/logo2.ba10b4af03869e69115ce84380e980aa.png"
+          alt="Keyman Logo" />
+        <h1>Predictive Text Studio</h1>
+      </div>
+      <div class="languages__container--actions">
+        <button
+          on:click={() => handleClick('information')}
+          class={selectedButton === 'information' ? 'selected' : 'deselected'}>
```
IanSeng: **[Suggestion]**: love the `===` and ternary operator. I would also avoid magic string on the tamplate

src/app/Languages.svelte
```diff
@@ -0,0 +1,212 @@
...
+          on:click={() => handleClick('sources')}
+          class={selectedButton === 'sources' ? 'selected' : 'deselected'}>
+          Sources
+        </button>
+        <button on:click={handleDownload} class="download"> Download </button>
+      </div>
+      <div class="languages__container--content">
+        {#if selectedButton === 'information'}
+          <LanguageInfo properties={languageInformation.properties} />
+        {:else if selectedButton === 'sources'}
```
IanSeng: **[Question]**: not very experience with svelte but if there is a way to clean up the logic here ?

IanSeng (commented):
it seems like this draft pr has a dependency with the task that I am working on, which is to find out a UI library. I think this pr would be a great team discussion for our next steps

src/app/Components/LanguageInfo.svelte
eddieantonio: @IanSeng could you clarify what you mean by this comment? I'm not sure I understand.

src/app/Languages.svelte
eddieantonio: There's a better way with svelte!

```suggestion
          class:selected={selectedButton === 'information'}>
```

src/app/Languages.svelte
eddieantonio: I think this falls under the resposibility of a router. See #48

src/app/Components/LanguageSources.svelte
eddieantonio: @IanSeng I feel like this is a legitemate use of a table? I'm really confused by what you mean that it makes it harder to _add_ accessibility functions? Tabular content (e.g., header + multiple rows) is accessible if using the correct table markup — try it in a screen reader!

src/app/Components/LanguageSources.svelte
```diff
@@ -0,0 +1,60 @@
...
+    <tr>
+      <th>Name</th>
+      <th>Size</th>
+      <th>Type</th>
+      <th>Actions</th>
+    </tr>
```
eddieantonio: Excellent use of `<th>`! I would also  use the [`<thead>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/thead) element here.

```suggestion
    <thead>
    <tr>
        <th>Name</th>
        <th>Size</th>
        <th>Type</th>
        <th>Actions</th>
      </tr>
    </thead>
```

src/app/Components/LanguageInfo.svelte
eddieantonio: Also note that Svelte uses [_scoped-styles_](https://svelte.dev/tutorial/styling) (which, honestly, I'm not a huge fan of), so these selectors do not actually affect anything outside of this `.svelte` element!

The magic sauce?

Svelte compiles this the following selector:

```css
img.svelte-167uv08 /*or some other similar unique ID */ {
  width: auto;
  height: 200px;
}
```

Although this scopes the style just for use in this component, it does have the side-effect of increassing the specificity of this selector; whereas the selector as written has a specificity of (0, 0, 1), the selector that Svelte generates has a specificity of (0, 1, 1).

---

Specificity: ignoring `!important` rules, selectors specificity has three components:
 - IDs `#like-this`
 - classes, pseudo-classes, and attributes `.like-this, [or=this], :hover`
 - elements, pseudo-elements `h1, ::before`

IanSeng: let me put my thought into a better sentence hahaha. What I meant was, straight up staying the `img` might not be a best practice.

eddieantonio (commented):
Oh lawd, I forgot to submit my comments yesterday! Okay, still no thorough review, as this is a draft. @pranavnarang let me know what you need to get this PR ready to review! 👍

src/app/Languages.svelte
pranavnarang: We can also have a list of action button names like:

```javascript
let buttons = ["information", "sources"];
```

And then we can loop over them to create the action buttons:

```javascript
{#each buttons as buttonName}

    <button 
        class:selected={buttonName===selectedButton}
        ...
    >
        {buttonName}
    </button>

{/each}
```

This might be a little cleaner and more scalable?

src/app/Languages.svelte
pranavnarang: I'm not sure if this falls under routing as the way the components are structured now, this is under one page/path but we can discuss this further. 

To clean up the logic we can map each button name to a component and have a function that returns the appropriate component by passing `selectedButton` as a parameter.

src/app/Components/Button.svelte
pranavnarang: Great point, Eddie has already setup some global css variables - was planning on adding these to that list when I pull from production.

src/app/Components/LanguageAvatar.svelte
pranavnarang: Thanks for mentioning this. In the `Sidebar` component I loop through each language and show an avatar.

```javascript
  {#each languages as language}
    <LanguageAvatar name={language.name} {selectedLanguage} />
  {/each}
```

 This basically determines if the `selectedLanguage` is the same as the `name` of the language in that avatar so that it can be styled accordingly.

src/app/Components/LanguageAvatar.svelte
pranavnarang: Would love to discuss what you've used as workarounds for magic strings for styling as I'm not too sure how to refactor this.

Would you add another variable like this?

```typescript
let selectedClassName: string = " selected";
```

And then change line 8 to:
```typescript
className = className + selectedClassName; 
```

eddieantonio: In Svelte I would **not** use strings like this for class names, since there's a better way! 

Remove the `className` variable entirely, then in where you need to use the markup, do the following:

```svelte
<div class="avatar" class:selected>
    ...
</div>
```

**EDIT**: please see: https://svelte.dev/tutorial/class-shorthand

pranavnarang: I'll look into this, thanks for the resource!

IanSeng (approved):
👍

src/app/App.svelte
```diff
@@ -1,20 +1,19 @@
...
 <main>
- 
   <Router {url}>
-    <nav>
+    <!-- <nav>
       <Link to="/">Landing</Link>
-      <Link to="testing">Testing Page</Link>
-    </nav>
+      <Link to="languages">Languages</Link>
+    </nav> -->
```
eddieantonio: Please just delete code instead of commenting it out!

src/app/Components/LanguageAvatar.svelte
```diff
@@ -0,0 +1,61 @@
+<script lang="ts">
+  export let name: string;
+  export let selectedLanguage: string = "";
+
+  export let isSelected: boolean = name === selectedLanguage;
+  export let className: string = "avatar";
+  if (isSelected) {
+    className = className + " selected";
+  }
```
eddieantonio:
```suggestion
  export let className: string = "avatar";
```

Additionally, is the `className` supposed to be modifiable from the outside? If not, then I would remove `export` as well:

```suggestion
  let className: string = "avatar";
```

src/app/Components/LanguageAvatar.svelte
eddieantonio:
```suggestion
    font-family: var(--main-font), sans-serif;
```

src/app/Components/Button.svelte
```diff
@@ -0,0 +1,122 @@
...
+   * @return {void}
+   */
+  const createStyles = (): void => {
+    className += color;
+    className += " " + size;
+    isOutlined && (className += " outline");
+    hasDropShadow && (className += " shadow");
+  };
+
+  beforeUpdate(async () => {
```
eddieantonio: Why is this async?

```suggestion
  beforeUpdate(() => {
```

src/app/Components/Button.svelte
```diff
@@ -0,0 +1,122 @@
...
+    flex-direction: column;
+    align-items: center;
+  }
+</style>
+
+{#if text && callbackFunction}
+  <div class="button-layout">
+    <button on:click={callbackFunction} class={className}> {text} </button>
+    {#if subtext}
+      <p class:subtext>{subtext}</p>
```
eddieantonio: This is an interesting use of the `class:subtext` feature! So here, it's _technically_ used wrong, but it has the right effect, haha!

So `class:subtext` applies the class "subtext" if the _variable_ `subtext` is truthy. That is, this is technically a conditional and not setting the class directly to subtext. This works here because of the `{#if subtext}` but I'd prefer just using the regular notation in this case (since I'm assuming the class _isn't_ conditional!):

```suggestion
      <p class="subtext">{subtext}</p>
```

src/app/Components/Button.svelte
```diff
@@ -0,0 +1,122 @@
...
+  import { beforeUpdate } from "svelte";
+
+  // props
+  export let color: string = "grey"; // optional
+  export let isOutlined: boolean = false; // optional
+  export let hasDropShadow: boolean = false; // optional
+  export let size: string = "medium"; // optional
+  export let text: string = ""; // required
+  export let subtext: string = ""; // optional
+  export let callbackFunction: any = null; // required
```
eddieantonio: Since consumers of this component will use need to set this property, I would rename it to what it means:

```suggestion
  export let onClick: (ev: MouseEvent) => void;
```

That way, users of this will use it like this:

```svelte
<Button onClick={() => console.log("clicked")}>Click me!</Button>
```

src/app/Components/Button.svelte
```diff
@@ -0,0 +1,122 @@
...
+  .button-layout {
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+  }
+</style>
+
+{#if text && callbackFunction}
+  <div class="button-layout">
+    <button on:click={callbackFunction} class={className}> {text} </button>
```
eddieantonio: Instead of the `generateStyle()` function and `beforeUpdate()` callback, I would do this:

```suggestion
    <button
        on:click={callbackFunction}
        class="{color} {size}"
        class:shadow={hasDropShadow}
        class:outline={isOutlined}
      ><slot></slot></button>
```

I have also used `<slot></slot>` instead of `{text}` as this enables the following usage of the element:

```svelte
<script>
    let count = 0;
    const increaseCount = () => void ++count;
</script>

<Button color="blue" hasDropShadow onClick={increaseCount}>
    I have been clicked {count} time(s)!
</Button>
```

Please see all my suggestions combined here: https://svelte.dev/repl/a6da043623f946c78bc60b6f0d2053af?version=3.29.0

src/app/Components/LanguageAvatar.svelte
```diff
@@ -0,0 +1,61 @@
...
+    box-shadow: 0px 10px 20px rgba(0, 0, 0, 0.1);
+  }
+</style>
+
+<div
+  class={className}
+  on:click={() => {
+    handleClick(name);
+  }}>
+  {name.substring(0, 1)}
```
eddieantonio: I am **not** going to request changes for this one change, but... since we're dealing with the worlds' languages, not all of them can use `name.substring(0, 1)` to get the first _abstract character_. There are many reasons for this, but it includes _grapheme clusters_  and in some dumb cases, _surrogate pairs_. [I have an entire presentation devoted to these two issues](https://eddieantonio.ca/unicode-es6/#/).

src/app/Components/LanguageInfo.svelte
eddieantonio: Yeah, BEM doesn't really make sense if we're using scoped CSS, which is the default in Svelte. I don't see the need for any changes right now, but something to keep in mind.

src/app/Components/LanguageSources.svelte
```diff
@@ -0,0 +1,123 @@
+<script lang="ts">
+  import Button from "./Button.svelte";
+  export let sources: any[];
```
eddieantonio: Honestly, this interface needs to be in some file, probably in `src/common/...`, but this will do for now:

```suggestion
  interface DictionarySource {
      name: string;
      size: number;
      type: string;
  }

  export let sources: DictionarySource[];
```

src/app/pages/Languages.svelte
```diff
@@ -0,0 +1,198 @@
...
+        name: "secondary_dictionary.xlsx",
+        size: "198",
+        type: "excel",
+      },
+    ],
+  };
+
+  export let languages: object[] = [
+    {
+      id: "1",
```
eddieantonio: Why are the IDs integers as strings?

src/app/pages/Languages.svelte
```diff
@@ -0,0 +1,198 @@
...
+      },
+      {
+        name: "secondary_dictionary.xlsx",
+        size: "198",
+        type: "excel",
+      },
+    ],
+  };
+
+  export let languages: object[] = [
```
eddieantonio:
```suggestion
  interface LanguageProject {
     id: number;
     name: string;
  }
  export let languages: LanguageProject[] = [
```

src/app/pages/Languages.svelte
```diff
@@ -0,0 +1,198 @@
...
+        type: "excel",
+      },
+      {
+        name: "Kinship Terms",
+        size: "32",
+        type: "direct entry",
+      },
+      {
+        name: "secondary_dictionary.xlsx",
+        size: "198",
```
eddieantonio: Why are the number strings?

src/app/pages/Languages.svelte
```diff
@@ -0,0 +1,198 @@
...
+  import LanguageSources from "../Components/LanguageSources.svelte";
+  import Sidebar from "../Components/Sidebar.svelte";
+  import Button from "../Components/Button.svelte";
+
+  export let selectedButton: string = "information";
+  export let selectedLanguage: string = "Kwakwala";
+
+  // Mock language data object - this would be read from localstorage/db
+  export let languageInformation = {
+    word_count: "74231",
```
eddieantonio: Two things: please use camelCase for the property name: `wordCount`. Second, we can implement that dynamic word count functionality like this:

```suggestion
    get wordCount(): number {
        return languageInformation.sources.reduce((sum, source) =>  sum + Number(source.size), 0);
    },
```

src/app/pages/Languages.svelte
```diff
@@ -0,0 +1,198 @@
...
+          src="https://keyman.com/cdn/deploy/img/logo2.ba10b4af03869e69115ce84380e980aa.png"
+          alt="Keyman Logo" />
+        <h1>Predictive Text Studio</h1>
+      </div>
+      <div class="languages__container--actions">
+        <!-- <button
+          on:click={() => handleClick('information')}
+          class={selectedButton === 'information' ? 'selected' : 'deselected'}>
+          Information
+        </button> -->
```
eddieantonio:
```suggestion
```

src/app/pages/Languages.svelte
```diff
@@ -0,0 +1,198 @@
...
+          color={"grey"}
+          isOutlined={(selectedButton === 'information')}
+          text="Information"
+          callbackFunction={() => handleClick('information')}
+        />
+        <!-- <button
+          on:click={() => handleClick('sources')}
+          class={selectedButton === 'sources' ? 'selected' : 'deselected'}>
+          Sources
+        </button> -->
```
eddieantonio:
```suggestion
```

src/app/pages/Languages.svelte
```diff
@@ -0,0 +1,198 @@
...
+        <Button
+          color="grey"
+          isOutlined={(selectedButton === 'sources')}
+          text="Sources"
+          callbackFunction={()=> handleClick('sources')}
+        />
+        <!-- <button on:click={handleDownload} class="download"> Download </button> -->
```
eddieantonio:
```suggestion
```

eddieantonio (changes requested):
I'm excited to get this wonderful design merged!

I've made some suggestions based on using Svelte features to their fullest, and some making use of TypeScript features as well.

Above all else, I would like it if **commented out code is deleted**. Remember that, above all, you are writing code for other humans, and not the computer. Comments are for explaining _why_, when the code gets confusing. But commented-out code often leaves me asking _why_ more often, and serves as a distractor from the code that _is_ there.

The solution to commenting out code: Well-labelled commits! If you feel the urge to comment out code, please first:
 1. commit what you got
 2. delete the code in the next commit, preferably with a meassage like <q>remove: old way of doing X</q>

This is preferrable to using leaving old code, but in comments.

---

Other people who share a similar opinion:
 - https://kentcdodds.com/blog/please-dont-commit-commented-out-code
 - https://markhneedham.com/blog/2009/01/17/the-danger-of-commenting-out-code/
 - https://softwareengineering.stackexchange.com/a/190222/284069
 - https://softwareengineering.stackexchange.com/a/213685/284069

src/app/components/LanguageAvatar.svelte
```diff
@@ -0,0 +1,69 @@
...
+
+  /**
+   * Returns the first letter / character of the language.
+   * TODO: Replace with a method that handles abstract cases
+   *
+   * @return {String} - the first character of the language name
+   */
+  const getLanguageShorthand = (): String => {
+    return name.substring(0, 1);
+  };
```
eddieantonio: I really like this refactor! It has a good name: `getLanguageShorthand()` that clearly indicates its intention, and has a TODO note. I would change the TODO note, though:

```suggestion
  /**
   * Returns the first letter / character of the language.
   * TODO: Handle non-BMP cases like "𐐆" for "𐐆𐑍𐑀𐑊𐐮𐑇" ("English" in the Deseret alphabet)
   * TODO: Handle multiple character grapheme clusters like "k̓"
   *
   * @return {String} - the first character of the language name
   */
  const getLanguageShorthand = (): String => {
    return name.substring(0, 1);
  };
```

src/app/components/LanguageAvatar.svelte
```diff
@@ -0,0 +1,69 @@
...
+</style>
+
+<div
+  class={className}
+  class:selected
+  on:click={() => {
+    handleClick(name);
+  }}>
+  {getLanguageShorthand()}
+</div>
```
eddieantonio, line 71: I didn't notice this before, but this _should_ be an ordinary `<a>` link to goes to a new project page. Don't do anything now though, since I think that can be handled in a future PR :)

src/app/components/LanguageSources.svelte
```diff
@@ -0,0 +1,126 @@
...
+
+    {#each sources as source}
+      <tr>
+        <td>{source.name}</td>
+        <td>{source.size}</td>
+        <td>{source.type}</td>
+        <td>
+          <div class="table__row--actions">
+            <p on:click={handleEdit}>Edit</p>
+            <p on:click={handleDelete}>Delete</p>
```
eddieantonio: I don't want this PR to drag on for too long so just keep this in mind for a future PR, but these should absolutely be `<button>` elements. As `<p>` elements, they are not keyboard accessible (at least without adding `tabindex="0"`). Since these are actions, the better thing is to simply use `<button>` elements, as they are keyboard accessible by default, and will be read by a screen reader as "button: Edit" and "button: Delete" 😌

src/app/components/Sidebar.svelte
```diff
@@ -0,0 +1,23 @@
...
+    justify-content: flex-start;
+    align-items: center;
+  }
+</style>
+
+<div class="sidebar">
+  {#each languages as language}
+    <LanguageAvatar name={language.name} {selectedLanguage} />
+  {/each}
+</div>
```
eddieantonio:
```suggestion
<nav class="sidebar">
  {#each languages as language}
    <LanguageAvatar name={language.name} {selectedLanguage} />
  {/each}
</nav>
```

src/app/pages/Languages.svelte
```diff
@@ -0,0 +1,194 @@
...
+      id: 4,
+      name: "P",
+    },
+  ];
+
+  // In case we want word count to be dynamic
+  // let languageWordCount: int = 0;
+  // languageInformation.sources.map((source) => {
+  //   languageWordCount += Number(source.size);
+  // });
```
eddieantonio:
```suggestion
```

😂 😭 😭

src/app/pages/Languages.svelte
```diff
@@ -0,0 +1,194 @@
...
+      <Sidebar {languages} {selectedLanguage} />
+    </div>
+
+    <div class="languages__container">
+      <div class="languages__container--header">
+        <img
+          src="https://keyman.com/cdn/deploy/img/logo2.ba10b4af03869e69115ce84380e980aa.png"
+          alt="Keyman Logo" />
+        <h1>Predictive Text Studio</h1>
+      </div>
```
eddieantonio:
```suggestion
      <header class="languages__container--header">
        <img
          src="https://keyman.com/cdn/deploy/img/logo2.ba10b4af03869e69115ce84380e980aa.png"
          alt="Keyman Logo" />
        <h1>Predictive Text Studio</h1>
      </header>
```

eddieantonio (approved):
LGTM! Just a few minor tweaks, and you can hit the merge button! 🥳

PR #59, opened 2020-09-21
feat: create zip with files

eddieantonio:
Replaces `generateZip()` with `createZipWithFiles()`. This creates an `ArrayBuffer`, which is supported both in NodeJS and in the browser. Not to mention, is compatible with the DOM [File](https://developer.mozilla.org/en-US/docs/Web/API/File) interface.


jienius (approved):
LGTM

PR #60, opened 2020-09-21
[#41] feat: store dictionary source files in indexedDB

### LINKED ISSUE #41: Store dictionary sources in LocalStorage or IndexedDB

jienius:
[Please delete all explanatory text in square brackets prior to creating the pull request!]
[Write a short explanation of your changes]
Added in IndexedDB Access library Dexie.js
Added in basic functionalities to store files inside IndexedDB

[Link to an existing issue number]
Fixes #41 
 OR
Resolves #

### DIFF
--- a/src/worker/indexedDBAccess.ts
+++ b/src/worker/indexedDBAccess.ts
@@ -1,0 +1,43 @@
+import Dexie from "dexie";
+
+const DB_NAME = "dictionary_sources";
+const VERSION = 1;
+
+/**
+ * files Table Scehma
+ * +-------+
+ * | id    |
+ * +-------+
+ * | name  |
+ * +-------+
+ * | file  |
+ * +-------+
+ */
+const FILES_TABLE_NAME = "files";
+const FILES_TABLE_SCHEMA = "++id, name, file";
+
+interface IFile {
+  id?: number;
+  name: string;
+  file: File;
+}
+
+class IndexDBAccess extends Dexie {
+  files: Dexie.Table<IFile, number>;
+
+  constructor() {
+    super(DB_NAME);
+    this.version(VERSION).stores({ files: FILES_TABLE_SCHEMA });
+    this.files = this.table(FILES_TABLE_NAME);
+  }
+
+  saveFile(name: string, file: File) {
+    return indexedDB.transaction("readwrite", indexedDB.files, async () => {
+      await indexedDB.files.where("name").equals(name).delete();
+      await indexedDB.files.put({ name, file });
+    });
+  }
+}
+
+const indexedDB = new IndexDBAccess();
+export default indexedDB;
--- a/src/worker/indexedDBAccess.ts
+++ b/src/worker/indexedDBAccess.ts
@@ -5,25 +5,25 @@ const VERSION = 1;
 
 /**
  * files Table Scehma
- * +-------+
- * | id    |
- * +-------+
- * | name  |
- * +-------+
- * | file  |
- * +-------+
+ * +------------------+
+ * | id (primary key) |
+ * +------------------+
+ * | name             |
+ * +------------------+
+ * | file             |
+ * +------------------+
  */
 const FILES_TABLE_NAME = "files";
 const FILES_TABLE_SCHEMA = "++id, name, file";
 
-interface IFile {
+interface StoredFile {
   id?: number;
   name: string;
   file: File;
 }
 
 class IndexDBAccess extends Dexie {
-  files: Dexie.Table<IFile, number>;
+  files: Dexie.Table<StoredFile, number>;
 
   constructor() {
     super(DB_NAME);
--- a/src/worker/main.ts
+++ b/src/worker/main.ts
@@ -1,12 +1,12 @@
 import indexedDB from "./indexedDBAccess";
 
-export interface ISaveFileEventPayload {
+export interface SaveFileEventPayload {
   name: string;
   file: File;
 }
 
 const handleSaveFileEvent = (event: MessageEvent) => {
-  const payload = event.data as ISaveFileEventPayload;
+  const payload = event.data as SaveFileEventPayload;
   indexedDB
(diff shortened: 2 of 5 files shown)


src/worker/main.ts
```diff
@@ -1 +1,32 @@
...
+//   const payload = event.data as ISaveFileEventPayload;
+//   indexedDB
+//     .saveFile(payload.name, payload.file)
+//     .then(() => {
+//       postMessage("Save Success");
+//     })
+//     .catch(() => {
+//       postMessage("Save Failed");
+//     });
+// };
```
eddieantonio: Please do not commit commented out code — that's what the commit log is for!

```suggestion
```

src/worker/main.ts
```diff
@@ -1 +1,32 @@
-self.postMessage("hello, from worker");
+import indexedDB from "./indexedDBAccess";
+
+export interface ISaveFileEventPayload {
```
eddieantonio: Redundant `I` in the name.

```suggestion
export interface SaveFileEventPayload {
```

src/worker/indexedDBAccess.ts
```diff
@@ -0,0 +1,43 @@
...
+const FILES_TABLE_NAME = "files";
+const FILES_TABLE_SCHEMA = "++id, name, file";
+
+interface IFile {
+  id?: number;
+  name: string;
+  file: File;
+}
+
+class IndexDBAccess extends Dexie {
```
eddieantonio, line 25: A quick look through [Dexie's documentation](https://dexie.org/docs/Tutorial/Best-Practices) doesn't show using it by extending or using inheritence;

What advanatages does inheritence here have over composition (e.g., just using the `db` as a property)?

src/worker/indexedDBAccess.ts
```diff
@@ -0,0 +1,43 @@
...
+ * +-------+
+ * | name  |
+ * +-------+
+ * | file  |
+ * +-------+
+ */
+const FILES_TABLE_NAME = "files";
+const FILES_TABLE_SCHEMA = "++id, name, file";
+
+interface IFile {
```
eddieantonio:
```suggestion
interface StoredFile {
```

src/worker/main.ts
eddieantonio: Is this the same interface as in `indexedDBAccess`? Can these two interface definitions be merged?

src/worker/indexedDBAccess.ts
```diff
@@ -0,0 +1,43 @@
...
+  file: File;
+}
+
+class IndexDBAccess extends Dexie {
+  files: Dexie.Table<IFile, number>;
+
+  constructor() {
+    super(DB_NAME);
+    this.version(VERSION).stores({ files: FILES_TABLE_SCHEMA });
+    this.files = this.table(FILES_TABLE_NAME);
```
eddieantonio, line 31: Doesn't Dexie export all tables as properties like this anyway? e.g., can't you do `db.files.where(...)` or `db.files.put(...)` etc?

src/worker/main.ts
```diff
@@ -1 +1,32 @@
...
+  name: string;
+  file: File;
+}
+
+const handleSaveFileEvent = (event: MessageEvent) => {
+  const payload = event.data as ISaveFileEventPayload;
```
eddieantonio:
```suggestion
  const payload = event.data as SaveFileEventPayload;
```

src/worker/indexedDBAccess.ts
```diff
@@ -0,0 +1,43 @@
...
+const FILES_TABLE_SCHEMA = "++id, name, file";
+
+interface IFile {
+  id?: number;
+  name: string;
+  file: File;
+}
+
+class IndexDBAccess extends Dexie {
+  files: Dexie.Table<IFile, number>;
```
eddieantonio:
```suggestion
  files: Dexie.Table<StoredFile, number>;
```

src/worker/indexedDBAccess.ts
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * files Table Scehma
+ * +-------+
+ * | id    |
+ * +-------+
+ * | name  |
+ * +-------+
+ * | file  |
+ * +-------+
```
eddieantonio: Good diagram! Could you mark which one is the primary key (e.g., indexed)? I believe `id` is an autoincrement primary key, right?

eddieantonio (changes requested):
Good job finding Dexie! It seems really neat, and seems to provide a clean, intuitive interface to IndexedDB. That said, I have some concerns! 😂

Note: I know that IndexedDB itself likes declaring interfaces with an `I` prefix, but I think it is completely unnecessary in our own custom interfaces in TypeScript. After all, if the `I` is to remind us that it's an interface, the IDE and compiler can tell us that instead!

src/worker/indexedDBAccess.ts
jienius: I followed their recommendation when using Dexie with TypeScript: https://dexie.org/docs/Typescript
It provides better type safety and simpler code

eddieantonio: Ah, okay, makes sense! That functionality _is_ already provided, but we would just like TypeScript to know exactly what happens when we access the `db.files` attribute!

src/worker/indexedDBAccess.ts
eddieantonio: Ignore this! I got it :)

src/worker/main.ts
jienius: You can, but I don't want them to share the same interface as one of them is for DB access while the other one is related to UI event

src/worker/indexedDBAccess.ts
jienius: Yes

eddieantonio (approved):
LGTM

PR #61, opened 2020-09-21
[WIP] world's worst predictive text studio

eddieantonio:
This is what I meant by "World's worst predictive text studio". In all likelihood, this PR will not be merged, and that's okay :)

I wanted to try implementing the whole flow from start to finish, and encountered a few issues:

 * having a UI/worker communication protocol would be nice, with **shared** interfaces that explain what the various messages are
 * `readXLXSfile` is a bad library :(

I learned some other things, but I've been keeping myself up late, so I'm gonna post this draft, then go to bed!

### DIFF
--- a/src/app/Upload.svelte
+++ b/src/app/Upload.svelte
@@ -1,12 +1,19 @@
 <script lang="ts">
+  /**
+   * NOTE: while we wait for more UI work to finish, this is a god class (or,
+   * uh, god component).
+   *
+   * Refactoring is necessary!
+   */
+
   const UPLOAD_INPUT_ID = "upload-input";
 
+  /** Whether the user is currently dragging something over this component: */
   let onDraggedOver = false;
-  let files = new Map<String, File>();
+  /** List of files to upload. */
+  let files = new Map<string, File>();
 
-  /**
-   * TODO: this MUST be moved to a new file (bad coupling).
-   */
+  /* TODO: this MUST be moved to a new file (bad coupling).  */
   let generatedCode: string | null = null;
 
   const handleDrop = (event: DragEvent) => {
@@ -21,17 +28,16 @@
         if (item.kind === "file") {
           const file = item.getAsFile();
           if (file !== null) {
-            files.set(file.name, file);
+            uploadFile(file);
           }
         }
       }
     } else {
       // Use DataTransfer interface to access the file(s)
       for (let file of event.dataTransfer.files) {
-        files.set(file.name, file);
+        uploadFile(file);
       }
     }
-    alert("File(s) dropped: " + files.size);
   };
 
   const handleDragOver = () => {
@@ -46,11 +52,17 @@
     const input = event.target as HTMLInputElement;
     if (input !== null && input.files) {
       for (let file of input.files) {
-        files.set(file.name, file);
-        alert(file.name);
+        uploadFile(file);
       }
     }
   };
+
+  function uploadFile(file: File) {
+    files.set(file.name, file);
+    /* Useless assignment so that Svelte picks it up: */
+    // See: https://svelte.dev/tutorial/updating-arrays-and-objects
+    files = files;
+  }
 </script>
 
 <style>
@@ -104,12 +116,22 @@
   on:drop|preventDefault={handleDrop}
   on:dragover|preventDefault={handleDragOver}
   on:dragleave|preventDefault={handleDragLeave}>
-  <img role="presentation" src="icons/upload-solid.svg" alt="" />
+  <img role="presentation" src="icons/upload-solid.svg" alt=""
+/>
   <span>Drag and drop here</span>
   <span>or</span>
   <label for={UPLOAD_INPUT_ID} class="upload-btn">Browse file</label>
-  <input id={UPLOAD_INPUT_ID} type="file" on:change={handleChange} />
+  <input id={UPLOAD_INPUT_ID} type="file" on:change={handleChange} data-cy="upload-file" />
 </div>
+
+<!--
+  TODO: remove this once https://github.com/eddieantonio/predictive-text-studio/pull/58 is merged.
+-->
+{#if files.size > 0}
+<button data-cy="download-kmp">
+  Download .kmp file
+</button>
+{/if}
 
 <!--
   TODO: remove this once https://github.com/eddieantonio/predictive-text-studio/pull/58 is merged.
--- a/src/app/Upload.svelte
+++ b/src/app/Upload.svelte
@@ -4,6 +4,11 @@
   let onDraggedOver = false;
   let files = new Map<String, File>();
 
+  /**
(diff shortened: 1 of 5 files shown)


src/app/Upload.svelte
```diff
@@ -16,17 +29,16 @@
         if (item.kind === "file") {
```
IanSeng, line 29: **[Suggestion]**: for the reason of clean coding, magic string should be avoided as much as possible. I would use enum or create a constant in this case

IanSeng, line 29: **[Suggestion]**: This nested logic is not very readable and can be done by using a `&` logic. I would also use ternary operator to further clean up

src/worker/main.ts
```diff
@@ -1 +1,66 @@
...
+
+/**
+ * A "fake" type; it's a string, but its contents MUST be a valid BCP-47 tag.
+ */
+type BCP47 = string & { __bcp46__: true };
+
+/**
+ * A model ID, according to spec.
+ */
+interface ModelID {
```
IanSeng, line 14: **[Suggestion]**: store interfaces in a different file

src/worker/main.ts
```diff
@@ -1 +1,66 @@
...
+      })
+      .catch((e) => {
+        self.postMessage({ error: e });
+      });
+  }
+};
+
+async function compileModel(spec: CompileMessage): Promise<string> {
+  const v = Object.entries(spec.files)[0];
+  const source = await parseFile(v[0], v[1]);
```
IanSeng, line 40: **[Question]**: if `await`is used here this function is essentially a synchronous function and `async` can be removed

src/worker/main.ts
```diff
@@ -1 +1,66 @@
...
+
+  return compileModelFromLexicalModelSource({
```
IanSeng, line 42: **[Question]**: this return seems like an object to me rather than a Promise with a string type ?

src/worker/main.ts
```diff
@@ -1 +1,66 @@
...
+): Promise<WordListFromArray> {
+  // TODO: the xlsx library is bad and won't parse an ArrayBuffer :(
+  return new WordListFromArray(filename, [
+    ["TTE", 15322],
+    ["E", 14422],
+    ["SEN", 8600],
+  ]);
+}
+
+function isCompileMessage(data: unknown): data is CompileMessage {
```
IanSeng, line 60: **[Question]**: out of curiosity, I have not used `is` in the past, what is `is` doing in this case ?

src/app/Upload.svelte
```diff
@@ -99,9 +126,26 @@
...
   <label for={UPLOAD_INPUT_ID} class="upload-btn">Browse file</label>
-  <input id={UPLOAD_INPUT_ID} type="file" on:change={handleChange} />
+  <input id={UPLOAD_INPUT_ID} type="file" on:change={handleChange} data-cy="upload-file" />
 </div>
+
+<!--
+  TODO: remove this once https://github.com/eddieantonio/predictive-text-studio/pull/58 is merged.
+-->
+{#if files.size > 0}
+<button data-cy="download-kmp" on:click={workInProgressCreateModel}>
```
IanSeng, line 141: **[Question]**: if this is not a throw away code/testing piece, a test-id would allow easy testing

src/app/Upload.svelte
```diff
@@ -41,11 +53,26 @@
...
+    /* Useless assignment so that Svelte picks it up: */
+    // See: https://svelte.dev/tutorial/updating-arrays-and-objects
+    files = files;
+  }
+
+  /**
+   */
+  function workInProgressCreateModel() {
+    compileModel(files).then((sourceCode: string) => {
+      generatedCode = sourceCode;
```
IanSeng, line 72: **[Question]**: I not sure if we are going to leverage on pure function programming, but mutating outside scope is definitely not a good practice, it harder to test and might be buggy

IanSeng (commented):
Although this is a draft pr, but I have dropped some comments for some of the things that can be improved

src/worker/main.ts
eddieantonio: [`await` is only syntactically valid in a function marked as `async`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await#:~:text=The%20await%20operator%20is%20used%20to%20wait%20for%20a%20Promise.%20It%20can%20only%20be%20used%20inside%20an%20async%20function.)

src/worker/main.ts
eddieantonio: `is` allows you to write a _dynamic_ function that asserts a _static_ type. So here, I wrote a function that checks whether the `unknown` thing is, in fact a `CompileMessage`.

See: https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards

src/app/Upload.svelte
eddieantonio: This is throw-away code!

But for the test-id, we're using `data-cy`, since that is [the official recommendation from Cypress](https://docs.cypress.io/guides/references/best-practices.html#Selecting-Elements).

src/app/Upload.svelte
eddieantonio: Haha, so because of Svelte's reactive assignments, it needs to be done. I agree, mutation should be avoided whenever possible, but Svelte has _indirect control-flow_ where [an assignment to a variable in the component's scope gets compiled to an `$$invalidate()` call](https://lihautan.com/compile-svelte-in-your-head-part-2/#invalidate), which updates the DOM.

eddieantonio:
@IanSeng so I should have marked this as experimental/proof-of-concept code, and now after seeing a few more pull requests, I _really_ don't think it should be merged. That said, I learned things, and will be making issues for them!

src/worker/main.ts
IanSeng: hahaha nice doc for my rusty async implementation

src/app/Upload.svelte
IanSeng: Hmmm, for the reason of maintainability and scaleability, `test-id` still seems to be better in the long run. For instance, if we scrape cypress off the project in the future, we would have to spend time in refactoring the name

src/app/Upload.svelte
IanSeng: awesome, time for me to dive deeper into learning Svelte

IanSeng:
> @IanSeng so I should have marked this as experimental/proof-of-concept code, and now after seeing a few more pull requests, I _really_ don't think it should be merged. That said, I learned things, and will be making issues for them!

I wasn't aware this is a draft/POC before making the comments but I have also learned a lot from this pr. 👍

src/app/Upload.svelte
eddieantonio: We're using `data-cy` in this project :)

eddieantonio:
Obsoleted by #78

PR #66, opened 2020-09-22
change: cleanup some placeholder code

eddieantonio:
I deleted a lot of placeholder code that was intended just to serve as an example for on-boarding. I also refactored how to access fixtures in the ava tests.


jienius (approved):
.

PR #69, opened 2020-09-22
test: store a file in IndexedDB

eddieantonio:
I refactored the Dexie-powered IndexedDB backend to make it easier to test (dependency-injection). This involves a fake IndexedDB that runs in Node.


src/worker/storage.ts
```diff
@@ -0,0 +1,48 @@
...
+  id?: number;
+  name: string;
+  contents: ArrayBuffer;
+}
+
+export class PredictiveTextStudioDexie extends Dexie {
+  files: Dexie.Table<StoredFile, number>;
+
+  constructor(options?: DexieOptions) {
+    super(DB_NAME, options);
```
jienius, line 29: Did Dexie put this options argument here intentionally for dependency injection?
And if I am reading this correctly you cannot read the stuff that is saved into that fake db, right?

jienius (approved):
.

src/worker/storage.ts
eddieantonio: Yeah! Dexie supports dependency injection, making it easier to test — no need to juggle globals!

And `fake-indexeddb` does NOT store things to disk, that's true, but things are stored in-process, which allows you to do this kind of assertion where you store something, and check to see if it's still there, as in:

https://github.com/eddieantonio/predictive-text-studio/blob/1dd06e8a0618027555812267a8f4bc7bb912d555/test/test-storage.ts#L13-L18

PR #70, opened 2020-09-23
Add download button

### LINKED ISSUE #32: Test uploading a sheet and downloading a KMP

xqiu1:
Waiting for the UI library
Question: Where to download the dictionary?

Fixes #32


eddieantonio (approved):
LGTM!

---

> Question: Where to download the dictionary?

I _think_ what will happen is this:

 1. files are sent to the worker
 2. the worker compiles all the files it knows about into a kmp
 3. the kmp bytes are sent back to the UI thread
 4. the UI thread uses [`URL.createObjectURL()`](https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL) to create something the user can actually download

PR #72, opened 2020-09-24
feat: generate a bare-minimal .kmp json

### LINKED ISSUE #51: Create a bare-minimum kmp.json file
This is the metadata file that is included in the generated `.kmp` zip archive.

A bare-minimum KMP must have the following:

```json
{
  "license": "mit", 
  "languages": ["crl", "crj"]
}
```

The license is ALWAYS `"mit"` 😂

And the languages is a list of supported languages (BCP-47), but for now, we should only support one.

Please see: https://help.keyman.com/developer/cloud/model_info/1.0/

jienius:
Fixes #51

### DIFF
--- a/src/worker/generate-modelinfo.ts
+++ b/src/worker/generate-modelinfo.ts
@@ -1,0 +1,46 @@
+/**
+ * Interfaces derived from the keyman documentation: https://help.keyman.com/developer/cloud/model_info/1.0/
+ */
+interface ModelInfo {
+  id?: string;
+  name?: string;
+  authorName?: string;
+  authorEmail?: string;
+  description?: string;
+  license: string;
+  languages?: string[];
+  lastModifiedDate?: string;
+  links?: ModelLinkInfo[];
+  packageFilename?: string;
+  packageFileSize?: number;
+  jsFilename?: string;
+  jsFileSize?: number;
+  isRTL?: boolean;
+  packageIncludes?: string;
+  version?: string;
+  minKeymanVersion?: string;
+  helpLink?: string;
+  sourcePath?: string;
+  related?: ModelRelatedInfo;
+}
+
+interface ModelLinkInfo {
+  name: string;
+  url: string;
+}
+
+interface ModelRelatedInfo {
+  deprecates?: boolean;
+  deprecatedBy?: boolean;
+  note?: string;
+}
+
+/**
+ * Given languages, generate a .kmp json file which contains metadata for .kmp zip archive.
+ */
+export function generateModelInfoJson(languages: string[]): string {
+  const license = "mit";
+  const modelInfo: ModelInfo = { license, languages };
+
+  return JSON.stringify(modelInfo);
+}
--- a/test/test-generate-modelinfo.ts
+++ b/test/test-generate-modelinfo.ts
@@ -1,0 +1,8 @@
+import test from "ava";
+
+import { generateModelInfoJson } from "@worker/generate-modelinfo";
+
+test("it should generate a JSON file with the given contents", async (t) => {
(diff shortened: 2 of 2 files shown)


src/worker/generate-metadata.ts
```diff
@@ -0,0 +1,18 @@
...
+ * Given languages, generate a .kmp json file which contains metadata for .kmp zip archive.
+ */
+export function generateMetadata(languages: string[]): string {
+  const license = "mit";
+  const metadata: Metadata = { license, languages };
+
+  return JSON.stringify(metadata);
+}
+
+export function generateMetadataJson(content: string): Blob {
```
jienius: Any recommendations on how to test Blob? I don't want to do weird juggling of converting it to ArrayBuffer or something

IanSeng: Unit test should be enough in this case. Since **generateMetadataJson** is just a function that return a value, I would isolate the dependencies by providing a mock string into content and test the return value.

src/worker/generate-metadata.ts
```diff
@@ -0,0 +1,18 @@
...
+  return new Blob([content], { type: "application/json" });
+}
```
eddieantonio: Because `createZipWithFiles()` takes in strings and not blobs, funnily enough, you don't need to even generate a blob to solve #51!

```suggestion
```

See: 
https://github.com/eddieantonio/predictive-text-studio/blob/494fde2eed2964f6315bb5d38f85dbdebc4deedf/src/worker/generate-zip.ts#L8-L18

test/test-generate-metadata.ts
```diff
@@ -0,0 +1,8 @@
+import test from "ava";
+
+import { generateMetadata } from "@worker/generate-metadata";
+
+test("it should generate a JSON file with the given contents", async (t) => {
+  const metadata = generateMetadata(["en", "zh"]);
+  t.is(metadata, `{"license":"mit","languages":["en","zh"]}`);
```
eddieantonio: 👍

src/worker/generate-metadata.ts
```diff
@@ -0,0 +1,18 @@
+interface Metadata {
+  license: string;
+  languages: string[];
+}
```
eddieantonio: Could you actually write an interface based on the docs here? https://help.keyman.com/developer/cloud/model_info/1.0/

Here's a formal schema: https://api.keyman.com/schemas/model_info.source.json?_ga=2.49804197.285811810.1600962912-50863687.1550158777

```suggestion
/**
 * See: https://help.keyman.com/developer/cloud/model_info/1.0/
 */
interface ModelInfo {
  license: string;
  languages: string[];
}
```

It seems that only license and languages are required properties; all other properties are optional!

eddieantonio (changes requested):
My biggest suggestion is to delete the blob generating code! :joy:
The other is to more fully specify the `ModelInfo` interface!

src/worker/generate-metadata.ts
jienius: > Because `createZipWithFiles()` takes in strings and not blobs, funnily enough, you don't need to even generate a blob to solve #51!
> 
> See:
> https://github.com/eddieantonio/predictive-text-studio/blob/494fde2eed2964f6315bb5d38f85dbdebc4deedf/src/worker/generate-zip.ts#L8-L18

But don't you need a blob to actually download it?

eddieantonio: Yes, but the blob is for the `.kmp` file (zip file) — not the JSON.

jienius: Okay, I see what you mean now

eddieantonio (approved):
LGTM!

Issue #45, opened 2020-09-19
Investigate a reusable frontend library


eddieantonio:
We don't have a lot of components, so we must judge whether it will legit makes our lives easier :D

PR #77, opened 2020-09-25
For Demo Purposes 

IanSeng:
[Please delete all explanatory text in square brackets prior to creating the pull request!]
[Write a short explanation of your changes]

[Link to an existing issue number]
Fixes #
 OR
Resolves #


eddieantonio:
Okay! So I'm going to try using both of these myself in my own tech spike. But honestly, if they don't add that much more value than what is already implemented in #58, then I think we'll pass on these component libraries. I appreciate the effort in investigating them and getting them setup. You can basically mark #45 as resolved! 😄

IanSeng:
> Okay! So I'm going to try using both of these myself in my own tech spike. But honestly, if they don't add that much more value than what is already implemented in #58, then I think we'll pass on these component libraries. I appreciate the effort in investigating them and getting them setup. You can basically mark #45 as resolved! 😄

I think this is a good call and thanks @eddieantonio  👍

Issue #53, opened 2020-09-20
Send the langauge information and uploaded file to the web worker; web worker should send back compiled model code


eddieantonio:
Will need to use `worker.postMessage()` to send both the BCP-47 tag to the web worker; the web worker should then `self.postMessage()` the compiled model source code (see #28).

PR #78, opened 2020-09-25
add generate kmp function

### LINKED ISSUE #40: Generate .kmp file for the dictionary

xqiu1:
Fixes #40


eddieantonio:
Hey @Xuechunqiu, is this PR ready? There are a lot of unlinked parts. I feel like a function like `generateKmp()` is not enough of a contribution to close #40 :/

**Edit**: missed key word 🙃

eddieantonio:
This is now **ready for review**!

src/app/Upload.svelte
```diff
@@ -3,6 +3,7 @@
 
   let onDraggedOver = false;
   let files = new Map<String, File>();
+  let downloadURL = "";
```
IanSeng:
```suggestion
  let downloadURL:  string;
```
**[Suggestion]**: If this string is not mutable I would use constant

src/app/Upload.svelte
```diff
@@ -100,6 +105,11 @@
     color: var(--blue);
     padding-bottom: 1em;
   }
+
+  .download.download--disabled {
```
IanSeng: **[Suggestion]**: I would refractor the class name to avoid confusion

src/app/Upload.svelte
```diff
@@ -113,4 +123,10 @@
   <span>or</span>
   <label for={UPLOAD_INPUT_ID} class="upload-btn">Browse file</label>
   <input id={UPLOAD_INPUT_ID} type="file" on:change={handleChange} />
+  <a
+    href={downloadURL ? downloadURL : "#"}
+    download="Example.kmp"
+    class="download"
```
IanSeng:
```suggestion
    class="download-link"
```
**[Suggestion]**: It's great to have more explicit class name, which it will later alleviate your testing

src/worker/generate-kmp.ts
```diff
@@ -0,0 +1,37 @@
...
+  const modelFile = compileModelFromLexicalModelSource({
+    format: "trie-1.0",
+    sources: sources,
+  });
+
+  const kmpFile = createZipWithFiles({
+    [`${modelId}.model.js`]: modelFile,
+    "kmp.json": kmpJsonFile,
+  });
+  return kmpFile;
```
IanSeng, line 40: **[Question]**: It seems like the linting is not doing the job here 😭, I will double check later

test/test-generate-kmp.ts
```diff
@@ -0,0 +1,29 @@
...
+  new WordListFromArray("wordlist.xlsx", [
+    ["TŦE", 13644],
+    ["E", 9134],
+    ["SEN", 4816],
+  ]),
+];
+const modelID = "nrc.en.mtnt";
+
+test("it should generate a kmp file", async (t) => {
+  const kmp = await generateKmp(langName, bcp47Tag, sources, modelID);
```
IanSeng: **[Question]**: is there a different way in Ava to test async function ?

test/test-link-storage-to-kmp.ts
```diff
@@ -0,0 +1,17 @@
...
+import * as IDBKeyRange from "fake-indexeddb/lib/FDBKeyRange";
+
+test("a function link the storage file to generate the kmp package", async (t) => {
+  const db = new PredictiveTextStudioDexie({
+    indexedDB: new FDBFactory(),
+    IDBKeyRange,
+  });
+  const storage = new Storage(db);
+
+  const kmpFile = await linkStorageToKmp(storage);
```
IanSeng, line 14: same as above

IanSeng (changes requested):
Great pr. I have some questions and a few suggestions

src/app/Upload.svelte
eddieantonio: `downloadURL` is mutated below (Svelte compiles the mutation to an `$$invalidate()` call).

src/app/Upload.svelte
eddieantonio: What would you name it?

src/worker/generate-kmp.ts
eddieantonio: What do you mean? I see absolutely nothing wrong with the code style here. Could you please clarify?

test/test-generate-kmp.ts
eddieantonio: A different way how so? I'm having difficulting understanding exactly what you mean by this

src/app/Upload.svelte
eddieantonio: I agree, `.download-link` is a better name, but never use a CSS class name in testing! 💀 😭 😭

src/app/Upload.svelte
eddieantonio: Hey @IanSeng. Thanks for the review comments. I've thought about this, and I notice many of your review comments are more about style than content. In this case, you would have noticed that `downloadURL` is mutable and cannot be turned into a constant by this line:

https://github.com/eddieantonio/predictive-text-studio/pull/78/files#diff-ce0dbfe92e42d16b2e8ce7fc58fdae2fR58

What's more, we have an `eslint` rule that would fail the build if we declare an identifier as `let` when it is never reassigned, so this particular comment would have already been caught by `eslint` before getting to this point :)

Please be more thorough in code review, and attempt to comment more on _content_ and _structure_ and less on _style_. If you have an issue with the style of the PR, take that up with our eslint and svelte-check rules! That's what they're there for. Let them do the talking :) Please review with concerns that only a human can check. Thanks!

**Edit**: "reassigned" instead of "assigned"

src/app/Upload.svelte
eddieantonio: @IanSeng for next time, may you please _suggest_ a better name? If a better name doesn't come to mind, either because the intent is not clear, or, you know, naming is hard, please the authors know exactly what the issue is (e.g., _why_ does this cause confusion) and that way we have an actionable way to respond to this review comment!

src/worker/link-storage-to-kmp.ts
```diff
@@ -0,0 +1,19 @@
+import { WordListFromArray } from "@predictive-text-studio/lexical-model-compiler";
+import { generateKmp } from "./generate-kmp";
+import Storage from "./storage";
+
+export async function linkStorageToKmp(storage: Storage): Promise<ArrayBuffer> {
+  const sourcesFromDB = await storage.fetchAllFiles();
+  const sources: WordListFromArray[] = [];
+  for (const { name, wordlist } of sourcesFromDB) {
+    sources.push(new WordListFromArray(name, wordlist));
+  }
```
pranavnarang: If I understand this correctly, these lines of code take each `StoredWordList` from the DB and convert it to a `WordListFromArray` to append to sources, which is later used to generate the kmp.

If so, I would probably move this to another helper function like:
```
let sources: WordListFromArray[] = generateSourcesFromStorage(storage);
``` 

to make `linkStorageToKmp` more readable/decouple functionality.

eddieantonio: @Xuechunqiu could you please factor out `generateSourcesFromStorage()` as Pranav suggested?

src/app/Upload.svelte
IanSeng: I am aware that some of the code is using id for testing and I thought this is what we have landed as a team, which is also why I put this comment here.

src/worker/generate-kmp.ts
IanSeng: There is inconsistency in the linting patter, for instance, there is no extra line after line 35. I wasn't sure about the linting rule we have and I thought that I might not implemented the husky hook correctly last time, which is why I put a "I will double check later here". Now that I am aware there is no rule in our linting.

test/test-generate-kmp.ts
IanSeng: I was using Jasmine that provide the `fakeAsync ` for the use of testing async function. Since I am extremely new to AVA I wasn't sure if Ava has something similar, which is why I put it as a **[Question]**

src/app/Upload.svelte
IanSeng: hmmm my apologies for the confusion. I was confused when I first look at this line and was wondering why `download--disabled` class is having the same CSS styling as `download`. That is why I think a better naming would be great and thought it could be a mistake in the naming or my mistake.

src/app/Upload.svelte
IanSeng: Hey @eddieantonio  I can definitely see you point and I hope that I do not come across as rude. Since we are working on a project as a team and I personally think it would be valuable to have consistent style/naming convention for the reason of maintainability and code readability. I can totally understand your perspective and will be more than happy take your feedback. Thank you.

src/worker/generate-kmp.ts
eddieantonio: It's irrelevant if the husky hook runs correctly, because the build literally fails if the linter dislikes it.

For future reviews, please do not focus on code style. That's why we're running the linter on CI. And it [does work](https://github.com/eddieantonio/predictive-text-studio/runs/1196158848). If the linter accepts it, then it's acceptable code, as far as style is concerned. Please comment on structure, form, and meaning instead.

test/test-generate-kmp.ts
eddieantonio: Ah, no, this is [the idiomatic way to write async tests](https://github.com/avajs/ava/blob/master/docs/01-writing-tests.md#async-function-support)!

IanSeng (approved):
After reading the feedback from @eddieantonio, I might be giving away too much of comments on granular styling and have removed the block.

src/worker/generate-kmp.ts
IanSeng: Thank you again for feedback and clarifying the norm for pr review for the project. I am more than happy to take your feedback.

src/app/Upload.svelte
eddieantonio: Confusion is okay! A helpful comment is:

> why [does] `download--disabled` class [have] the same CSS styling as `download`?

And you could add:

> Could you make a clearer name for "download"?

Also note, that I'm using BEM here, so it's `.download` with the `--disabled` modifier/state; although in all fairness, `.download` is a redundant selector (increases the specificity), so now that I think about it, I can safely remove the `.download` part of the selector! :)

src/worker/generate-kmp.ts
```diff
@@ -0,0 +1,38 @@
...
+ * @param modelId
+ */
+export async function generateKmp(
+  langName: string,
+  bcp47Tag: string,
+  sources: WordListFromArray[],
+  modelId: string
+): Promise<ArrayBuffer> {
+  const kmpJsonFile = generateKmpJson({
+    languages: [{ name: langName, id: bcp47Tag }],
```
eddieantonio, line 26: @Xuechunqiu a thing we didn't notice earlier is that the `kmp.json` file needs to have the same modeld as in the `.kmp` file; however, since we didn't provide it as an argument here, the name in the `kmp.json` may not match the filename here.

Since `generateKmpJson()` automatically generates its own model ID, this **will** get out of sync, and will not install in Keyman (I tried it!). The real solution is to change `generateKmpJson()` so that it accepts a model ID and does not generate it itself, however, this will make it work in the meantime:

```suggestion
    languages: [{ name: langName, id: bcp47Tag }],
    // TODO: change generateKmpJson() to accept a modelID argument instead of splitting it here:
    authorID: modelId.split(".")[0],
    modelTag: modelId.split(".")[2],
```

eddieantonio (changes requested):
Hey @Xuechunqiu, just one quick change we forgot about, and then LGTM! I would appreciate a review from @jienius before merging though, as much of what is covered here conflicts with #53, and will need to be reconciled before merging both.

src/app/Upload.svelte
IanSeng: Nice one. Thank you

src/worker/generate-kmp.ts
xqiu1: I see what you mean. In addition to these quick changes, do I need to change generateKmpJson() for now, or do it later?

eddieantonio (approved):
LGTM

src/worker/generate-kmp.ts
eddieantonio: I think that's best kept for a seperate PR :)

eddieantonio: Update: it's issue #91 if you want to tackle it later

src/app/Upload.svelte
jienius: URL or Url, HTTP or Http, HTML or Html

test/test-generate-kmp.ts
```diff
@@ -0,0 +1,29 @@
...
+const langName = "English";
+const bcp47Tag = "en";
+const sources = [
+  new WordListFromArray("wordlist.xlsx", [
+    ["TŦE", 13644],
+    ["E", 9134],
+    ["SEN", 4816],
+  ]),
+];
+const modelID = "nrc.en.mtnt";
```
jienius: ID or Id

src/app/Upload.svelte
```diff
@@ -100,6 +105,16 @@
...
     padding-bottom: 1em;
   }
+
+  .download-link {
+    display: block;
+    margin-top: 1.5em;
+  }
+
+  .download-link--disabled {
+    color: #666;
```
jienius: I would extract this color into the global style since disabled color will be potentially used in multiple places

test/test-generate-kmp.ts
xqiu1: I will use Id to make the variable name consistent.

src/app/Upload.svelte
eddieantonio: It seems our convension is the following:

```suggestion
  let downloadUrl = "";
```

Needs to be changed in the rest of the file, though!

src/app/Upload.svelte
xqiu1: already changed :)

jienius (approved):
LGTM! Good job!

PR #79, opened 2020-09-26
[#73] Store interfaces that are shared between app and worker elsewhere

### LINKED ISSUE #73: Store common interfaces in a common file —accessible both to the app and worker bundles
There are a few interfaces that should be accessible both to the worker and the app alike. They should both access these interfaces.

For example, the `ModelID` interface is useful both for the worker and the app, both bundles should know about this:

```typescript
interface ModelID {
  author: string;
  bcp47: BCP47;
  uniq: string;
}
```

jienius:
Fixes #73

### DIFF
--- a/src/common/events.ts
+++ b/src/common/events.ts
@@ -1,0 +1,4 @@
+export interface SaveFileEventPayload {
+  name: string;
+  file: File;
+}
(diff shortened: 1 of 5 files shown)


eddieantonio (approved):
This is awesome, and just what I imagined it working like!

L(very)GTM!

PR #81, opened 2020-09-27
change: generate kmp.json from relevant options

eddieantonio:
Changes `generateKmpJson()` to take in options relevant to Predictive Text Studio. Any options in `kmp.json` that can be intelligently inferred are automatically generated. There are also some weird things in the `kmp.json` schema that just have to be a certain way, so this function generates the things that Keyman wants :)


xqiu1 (approved):
LGTM

PR #82, opened 2020-09-28
remove: custom BCP-47 types

eddieantonio:
I somehow didn't notice that we already had types for this library, so I deleted the ones I wrote


jienius (approved):
.

Issue #74, opened 2020-09-24
Establish a worker/app communication protocol


eddieantonio:
The app (UI thread) should instantiate the worker once; then, the app and the worker can communicate via `postMessage()`. But _how_ should they communicate. Here's a proposal:

Use [discriminating unions](https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#discriminating-unions) to establish messages and their payload.

We first list the different kinds of messages that can be sent:

```typescript
type Message = UploadFileMessage | CompilationCompleteMessage;
```

We will add many, _many_ more messages!

These are interfaces that both the worker and app know (see #73).

All of these interfaces have one **common property**. Let's call it `method`:

```
interface UploadFileMessage {
    method: "upload-file";

    filename: string;
    rawData: ArrayBuffer;
}

interface CompilationCompleteMessage {
    method: "compiliation-complete";

    downloadURL: string;
}
```

---

Usage:

In both the app and the worker, we send one of these messages to `postMesage()`:

```typescript
// in the app
worker.postMessage({
  method: "upload-file",
  filename: "example.xlsx",
  rawData: await file.arrayBuffer(),
});

// in the worker
self.postMessage({
   method: "compilation-complete",
   downloadURL: URL.createObjectURL(kmpFile),
});
```

Related to #73

eddieantonio:
Hey @IanSeng, any progress on this? I'm realizing that #53 touches upon this, and we might need to make use of the work in this issue over in that PR.

If you haven't started, could you please unassign yourself, and move this to the "TODO" column?

IanSeng:
> Hey @IanSeng, any progress on this? I'm realizing that #53 touches upon this, and we might need to make use of the work in this issue over in that PR.
> 
> If you haven't started, could you please unassign yourself, and move this to the "TODO" column?

Apologies for the delayed response, just got email notified. Since I haven't started working on this task and there is a dependency with #53, I will unassign myself for this one and pick up another.

eddieantonio:
@jienius, I think #83 now actually covers _this_ issue instead of #53!

eddieantonio:
I've assigned this to @jienius as #83 is essentially solving this issue! 😂

eddieantonio:
Resolved in #94 🥳

PR #83, opened 2020-09-28
[#53] Compile model in work and return to app

### LINKED ISSUE #53: Send the langauge information and uploaded file to the web worker; web worker should send back compiled model code

jienius:
Fixes #53


src/common/events.ts
```diff
@@ -1,4 +1,25 @@
...
+}
+
+export interface CompileModelRequestPayload extends EventPayload {
   name: string;
   file: File;
 }
+
+export interface CompileModelResponsePayload extends EventPayload {
+  code?: string;
+}
```
eddieantonio: I would use a [discriminating union](https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#discriminating-unions) here instead.

Why?

 1. use a **self-documenting string** as the `type` property that is visible while debugging!
 2. TypeScript's type inferencing algorithm understands discriminating unions when you use them in a **switch** or **if**. That means, **no more `if (event.type === BLAH) { payload as BlahPayload }`. TypeScript already _infers_ that it's a `BlahPayload`!


```suggestion
export type AppWorkerEvent =
  | SaveFileRequest
  | CompileModelRequest
  | CompileModelResponse;

interface EventPayload {
  errorMessage?: string;
}

export interface SaveFileRequest extends EventPayload {
  type: "save-file";
  name: string;
  file: File;
}

export interface CompileModelRequest extends EventPayload {
  type: "compile-model";
  name: string;
  file: File;
}

export interface CompileModelResponse extends EventPayload {
  type: "compilation-result";
  code?: string;
}
```

src/app/Upload.svelte
```diff
@@ -1,4 +1,12 @@
 <script lang="ts">
+  import type {
+    SaveFileRequestPayload,
+    EventPayload,
+    CompileModelRequestPayload,
+    CompileModelResponsePayload,
+  } from "../common/events";
```
eddieantonio: Can we make this an import to `@common/events` instead?

```suggestion
  } from "@common/events";
```

This will require a change to the `tsconfig.json` to include `@common/` in the `baseURL`/`paths` config. Please see the root `tsconfig.json` for details!

src/app/Upload.svelte
```diff
@@ -18,6 +26,7 @@
           if (file !== null) {
             files.set(file.name, file);
             saveToIndexedDB(file.name, file);
+            compileModel(file.name, file);
```
eddieantonio: In general, I feel like a lot of this code can and SHOULD be refactored and extracted to its own standalone TypeScript file. I'd prefer to keep the `.svelte` code as dumb as possible, so I think it's about time to extract some methods!

src/app/Upload.svelte
```diff
@@ -44,14 +54,41 @@
       for (let file of input.files) {
         files.set(file.name, file);
         saveToIndexedDB(file.name, file);
+        compileModel(file.name, file);
```
eddieantonio: So what I'm understanding from these three duplicated lines is that a new model is compiled for _each_ uploaded file? This is incorrect, as one model is supposed to be compiled from **all** uploaded files.

Also, why are we storing the files in the IndexedDB if we're never retrieving them?

I feel like the correct solution is to ask the worker to compile the current model, and the worker retrieves all files stored, and compiles **one** model, which is then made available for download.

src/app/Upload.svelte
```diff
@@ -44,14 +54,41 @@
...
+      const response = event.data as EventPayload;
+      if (response.errorMessage) {
+        console.error(response.errorMessage);
+      }
+      worker.terminate();
+    };
+  };
+
+  const compileModel = (name: string, file: File) => {
+    const worker = new Worker("worker.js");
```
eddieantonio: As our code gets less "proof-of-concept" and more serious, I believe that the worker should be instantiated exactly _once_ by the app. So I believe this needs to be refactored and extracted too.

src/app/Upload.svelte
```diff
@@ -44,14 +54,41 @@
...
+    const request: CompileModelRequestPayload = {
+      type: EventType.CompileModelRequest,
+      name,
+      file,
+    };
+    worker.postMessage(request);
+    worker.onmessage = (event) => {
+      const response = event.data as CompileModelResponsePayload;
+      if (response.errorMessage) {
+        console.error(response.errorMessage);
```
eddieantonio: I feel like there should be some error reprted to the end-user, but I'm not sure how yet :/

src/worker/main.ts
```diff
@@ -1,24 +1,65 @@
...
-const handleSaveFileEvent = async (event: MessageEvent) => {
-  const payload = event.data as SaveFileEventPayload;
+const handleEvent = async (event: MessageEvent) => {
+  const payload = event.data as EventPayload;
+  switch (payload.type) {
+    case EventType.SaveFileRequest: {
+      await handleSaveFileEvent(payload as SaveFileRequestPayload);
+    }
+    case EventType.CompileModelRequest: {
+      await handleCompileModel(payload as CompileModelRequestPayload);
```
eddieantonio: A few things!

Every `switch/case` must have a `default`. Also, I don't assume you intended a fallthrough here, so I added break.
Why did you use `await` here? What purpose does it serve if this is a void function, and nothing `await`s `handleEvent()`'s return?

```suggestion
      handleSaveFileEvent(payload as SaveFileRequestPayload);
      break;
    }
    case EventType.CompileModelRequest: {
      handleCompileModel(payload as CompileModelRequestPayload);
      break;
    }
    default: {
       // TODO?????
```

src/worker/main.ts
```diff
@@ -1,24 +1,65 @@
...
   } catch (e) {
-    postMessage("Save Failed");
-    return;
+    const response: EventPayload = {
+      type: EventType.SaveFileResponse,
+      errorMessage: "Save Failed",
+    };
+    postMessage(response);
   }
+};
```
eddieantonio: This `try/catch` should not attempt catching the error from `postMessage()`. So use one `postMessage()` after the `try/catch`.

src/worker/main.ts
```diff
@@ -1,24 +1,65 @@
...
 
-  postMessage("Save Success");
+const handleCompileModel = async (payload: CompileModelRequestPayload) => {
+  try {
+    const buffer = await payload.file.arrayBuffer();
+    const wordlist = await readExcel(buffer);
```
eddieantonio: Not important, but these tasks can be done in parallel by using [`Promise.allSettled()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled)

src/worker/main.ts
```diff
@@ -1,24 +1,65 @@
...
+      type: EventType.CompileModelResponse,
+      code,
+    };
+    postMessage(response);
+  } catch (e) {
+    const response: CompileModelResponsePayload = {
+      type: EventType.CompileModelResponse,
+      errorMessage: e as string,
+    };
+    postMessage(response);
```
eddieantonio: Same thing about `postMessage()` from before!

src/app/Upload.svelte
```diff
@@ -44,14 +54,41 @@
...
+  const compileModel = (name: string, file: File) => {
+    const worker = new Worker("worker.js");
+    const request: CompileModelRequestPayload = {
+      type: EventType.CompileModelRequest,
+      name,
+      file,
+    };
+    worker.postMessage(request);
+    worker.onmessage = (event) => {
+      const response = event.data as CompileModelResponsePayload;
```
eddieantonio: I love how this receives the compiled model's code, and then just... throws it out 😆

eddieantonio (commented):
Thank you for taking time to do this! However, I'm honestly not sure about this :/

It kind of touches upon #74.

So we could accept the changes here, and then focus _exclusively_ on refactoring and tests for the few PRs.

I would also like to see tests for this. Does this actually... you know... work? 😆

So how would you like to continue @jienius? Would you like to get this merged and then focus our attention next to testing, documenting, and refactoring? Or would you rather get that over with right now?

I feel like we're dragging on our "proof-of-concept" period, and I'd like to see more "production-ready" code. This means tests, documentation, and better factoring!

src/worker/main.ts
jienius: > Thank you for taking time to do this! However, I'm honestly not sure about this :/
> 
> It kind of touches upon #74.
> 
> So we could accept the changes here, and then focus _exclusively_ on refactoring and tests for the few PRs.
> 
> I would also like to see tests for this. Does this actually... you know... work? 
> 
> So how would you like to continue @jienius? Would you like to get this merged and then focus our attention next to testing, documenting, and refactoring? Or would you rather get that over with right now?
> 
> I feel like we're dragging on our "proof-of-concept" period, and I'd like to see more "production-ready" code. This means tests, documentation, and better factoring!

I will try to address your comments and add more tests. I don't expect this PR to be merged any time soon tbh, there are just so many... uncertainty.

src/worker/main.ts
jienius: > Thank you for taking time to do this! However, I'm honestly not sure about this :/
> 
> It kind of touches upon #74.
> 
> So we could accept the changes here, and then focus _exclusively_ on refactoring and tests for the few PRs.
> 
> I would also like to see tests for this. Does this actually... you know... work? 
> 
> So how would you like to continue @jienius? Would you like to get this merged and then focus our attention next to testing, documenting, and refactoring? Or would you rather get that over with right now?
> 
> I feel like we're dragging on our "proof-of-concept" period, and I'd like to see more "production-ready" code. This means tests, documentation, and better factoring!

Cool

> Thank you for taking time to do this! However, I'm honestly not sure about this :/
> 
> It kind of touches upon #74.
> 
> So we could accept the changes here, and then focus _exclusively_ on refactoring and tests for the few PRs.
> 
> I would also like to see tests for this. Does this actually... you know... work? 
> 
> So how would you like to continue @jienius? Would you like to get this merged and then focus our attention next to testing, documenting, and refactoring? Or would you rather get that over with right now?
> 
> I feel like we're dragging on our "proof-of-concept" period, and I'd like to see more "production-ready" code. This means tests, documentation, and better factoring!

I will try to address your comments and add more tests. I don't expect this PR to be merged any time soon tbh, there are just so many... uncertainty.

jienius:
> Thank you for taking time to do this! However, I'm honestly not sure about this :/
> 
> It kind of touches upon #74.
> 
> So we could accept the changes here, and then focus _exclusively_ on refactoring and tests for the few PRs.
> 
> I would also like to see tests for this. Does this actually... you know... work? 
> 
> So how would you like to continue @jienius? Would you like to get this merged and then focus our attention next to testing, documenting, and refactoring? Or would you rather get that over with right now?
> 
> I feel like we're dragging on our "proof-of-concept" period, and I'd like to see more "production-ready" code. This means tests, documentation, and better factoring!

I will try to address your comments and add more tests. I don't expect this PR to be merged any time soon tbh, there are just so many... uncertainty.

src/worker/main.ts
jienius: I forgot it does fallthrough...

src/worker/main.ts
jienius: Why? readExcel is dependent on the buffer

eddieantonio: Oh derp, you're right, I didn't notice that >.<

jienius:
I am closing this PR as I have completely refactored my work which makes this PR is no longer relevant

PR #85, opened 2020-09-30
feat: Storage.fetchAllFiles()

### LINKED ISSUE #84: Implement `Storage.fetchAllFiles()` (retrieve files from the database!)
Our `Storage` class already stores files in the database, but we never retrieve them! So we should have a function that returns ALL of the files.  Since the intended use case of this is retrieving **word lists**, it's probably best if this fetches word lists! Something with this interface, maybe:

```TypeScript
type WordList = [/*word:*/ string, /*count:*/ number][];
async fetchAllFiles(): Promise<[string, WordList]>;
```

To officially close this issue, we must have tests that insure, if we store multiple files,and they all come back out as word lists!

eddieantonio:
Fixes #84

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -67,7 +67,7 @@ const workerConfiguration = {
   },
   plugins: [
     typescript({
-      tsconfig: "./tsconfig.json",
+      tsconfig: "src/worker/tsconfig.json",
     }),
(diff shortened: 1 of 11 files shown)


xqiu1 (approved):
LGTM

PR #88, opened 2020-10-01
fix: remove old hack

eddieantonio:
there was a brief moment of time when I had to cast something to unknown to get the TypeScript compiler to stop screaming at me; that time is no longer. No more gross casting to unknown!


jienius (approved):
.

pranavnarang (approved):
LGTM!

PR #89, opened 2020-10-01
refactor(test): factor out common setup code

eddieantonio:
Practically every test in test-storage needs its own, empty PredictiveTextStudio database, alongside a Storage object that uses it. This extracts that into a beforeEach hook so that the following tests are easier to write (less boilerplate).


jienius (approved):
.

pranavnarang (approved):
LGTM 👍

Issue #76, opened 2020-09-24
Avoid Svelte's scoped-styles and use BEM instead?


eddieantonio:
Although scoped CSS — writing `<style>` directly within a Svelte component — is convenient, it's not without its costs:

Styles distributed among multiple files — if the scoped styles are not factored well, this could end up being confusing.

Scoping is accomplished by adding a unique class to each element it affects; this has the side-effect of increasing the specificity of every selector written with a scoped `<style>` block.

---

**What conventions should we have for organizing CSS?**

I personally:
 - have my reservations about scoped CSS, but I can see its utility in _truly_ reusable components
 - like BEM naming (https://en.bem.info/methodology/quick-start/, https://css-tricks.com/bem-101/, https://9elements.com/bem-cheat-sheet/), but it can get quite "philosophical"!

Thoughts?

eddieantonio:
We'll use whatever works at the time and refactor when needed :D

PR #90, opened 2020-10-02
[#48] feat: Find an appropriate router for Svelte 

### LINKED ISSUE #48: Find an appropriate router for Svelte

IanSeng:
# Summary 
- Implemented routing feature into the application by using the library **svelte-routing**. 
- Restructured the folder into `src/pages` & `src/components` 
- Moved all content from `App.svelte` to` LandingPage.svelte` 
- Created a Testing page for the purpose of manual testing for the routing functionality
- Introduced a global css file

## Note: 
- This library was fully written in JS and I remove `lang="ts"` from the `App.svelte` to avoid the complain from TS. Other than  Sapper framework, most of the Svelte routing libraries I have found were written in JS; thus, this might seem to be inevitable. @eddieantonio  it would be great if I could have your insight on this. 

## Additional 
@pranavnarang  I wasn't aware of the links that you have provided initially at the comment of the task and the blog link comes in really handy for me to fix some of the issues, since the author didn't keep original doc on GitHub update. Kodus to your and thank you so much for the links <3 

## Issue 
Fixes #48


package.json
```diff
@@ -14,7 +14,7 @@
     "lint": "eslint '{src,test,cypress}/**/*.ts'",
     "lint:fix": "eslint --fix '{src,test,cypress}/**/*.ts'",
     "prettier": "prettier --write '{src,test,cypress}/**/*.ts'",
-    "start": "sirv public",
+    "start": "sirv public -s",
```
eddieantonio, line 17: Good stuff!

src/app/App.svelte
```diff
@@ -1,50 +1,17 @@
-<script lang="ts">
```
eddieantonio, line 1: This is totally acceptable if we only do routing in `App.svelte`

src/app/App.svelte
```diff
@@ -1,50 +1,17 @@
...
-  import Upload from "./Upload.svelte";
-  import BCP47Tag from "./BCP47Tag.svelte";
-  export let name: string;
+<script>
+  import { Router, Route } from "svelte-routing";
+  import LandingPage from "./pages/LandingPage.svelte";
+  import Testing from "./pages/Testing.svelte";
 
-  let files: any;
+  // Used for SSR. A falsy value is ignored by the Router.
```
eddieantonio, line 6: Good comment!

src/app/App.svelte
```diff
@@ -1,50 +1,17 @@
...
-  }
-
-  :global(h1, h2, h3, h4, h5, h6) {
-    font-family: var(--main-font), sans-serif;
-  }
-
-  :global(pre, code, samp) {
-    font-family: var(--mono-font), monospace;
-  }
-</style>
```
eddieantonio, line 38: Could you actually keep these **global** styles here in `App.svelte`? These definitions should be cascaded to _all_ pages in the app 👍 

---

Honestly, figuring out where out global styles _should_ be is an issue in and of itself — I'm not entirely happy with it being in the `.svelte` files. But that can be handled in a different PR 😅  See also: #76.

src/app/App.svelte
```diff
@@ -1,50 +1,17 @@
...
-  <p>Hi, my name is Xuechun!</p>
-  <p>Hi my name is Eddie!</p>
-  <p>Pranav</p>
-  <h2>Ian</h2>
-  <input type="file" data-testid="input-file" bind:files />
+  <Router {url}>
+    <div>
+      <Route path="/" component={LandingPage} />
+      <Route path="testing" component={Testing} />
+    </div>
```
eddieantonio: I checked out this branch and tried it on my local machine and found out that these extra `<div>` elements can safely removed — to my knowledge, they do not affect the functionality in any way whatsoever. 

I'm would like to avoid making the DOM tree any deeper, as a big, complicated DOM tree makes most sites sluggish  (I'm looking at you, new reddit.com!)

```suggestion
    <Route path="/" component={LandingPage} />
    <Route path="testing" component={Testing} />
```

server.js
```diff
@@ -0,0 +1,14 @@
...
+  const { html } = render({ url: req.url });
+
+  res.write(`
+    <!DOCTYPE html>
+    <div id="app">${html}</div>
+    <script src="/dist/bundle.js"></script>
+  `);
+
+  res.end();
+}).listen(3000);
```
eddieantonio: Why is this file necessary? As we're using static-hosting, we do _not_ have a server that we can deploy like this. Was this file added in mistake when initially testing things?

src/app/pages/LandingPage.svelte
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  :global(h1, h2, h3, h4, h5, h6) {
+    font-family: var(--main-font), sans-serif;
+  }
+
+  :global(pre, code, samp) {
+    font-family: var(--mono-font), monospace;
+  }
+</style>
```
eddieantonio: (see my comments in `App.svelte`)

```suggestion
```

src/app/main.ts
```diff
@@ -2,6 +2,7 @@ import App from "./App.svelte";
 
 const app = new App({
   target: document.body,
+  hydrate: true,
```
eddieantonio: If `server.js` is not necessary, then this line is not necessary. As far as I understand, hydratable components only make sense if we have server-side rendering, which we don't! 😆 

```suggestion
```

rollup.config.js
```diff
@@ -27,6 +27,7 @@ const appConfiguration = {
   },
   plugins: [
     svelte({
+      hydratable: true,
```
eddieantonio: (see my comment in `server.js` and `src/app.main.ts`)

```suggestion
```

eddieantonio (changes requested):
This PR brings some much welcomed improvements! I would just like a few redundant things removed, but other than that, LGTM!

I really like the restructuring of `app/pages` and `app/components`!

server.js
IanSeng: Without this file the reload function will not work on different route other than the main route. 

**You can test this by:**  
1. Delete this file and hydrate: true in both `main.ts` and `rollup.config `
2. Spin up the application 
3. Go to localhost:5000/testing 
4. Refresh the page and the error will show

src/app/main.ts
IanSeng: Same as above, this `hydrate: true` is needed to make the routing work as per the documentation from the author

rollup.config.js
IanSeng: same as above

src/app/App.svelte
IanSeng: let me try creating a global css and see how things work so that we can keep the `App.svelte` as minimal as possible

server.js
eddieantonio: I followed these steps but my local copy still works. I've spun up the application using `yarn run dev` which delegates to `sirv -s`. Either way, `server.js` would not work in Vercel, where we're deploying the application

src/app/App.svelte
IanSeng: Opps nice catch

server.js
IanSeng: here is the doc from the author, [https://github.com/EmilTholin/svelte-routing](https://github.com/EmilTholin/svelte-routing )

src/app/App.svelte
eddieantonio: Yeah, honestly, getting the global styles out of `App.svelte` is the better option 👍

server.js
eddieantonio: I see, and that usage scenario is _not_ applicable in this project. The main usage docs assume server-side rendering, but we don't (and can't) have that configured. Hence, this file remains unused.

IanSeng: Yes, you are correct and same for the comment bellow hydrate: true in both main.ts and rollup.config. But I am not sure why it wasn't working without without these config last night. But yeah since they are redundant, I will take them out.

eddieantonio (approved):
LGTM!

PR #94, opened 2020-10-03
[#53 #74]Compile model in work and return to app

### LINKED ISSUE #53: Send the langauge information and uploaded file to the web worker; web worker should send back compiled model code

jienius:
Fixes #53 #74 

I am opening a new PR as I have done a major refactor

I have found this amazing [library](https://github.com/GoogleChromeLabs/comlink) which makes web worker much more intuitive
I think it makes setting up actual web workers on nodejs unnecessary as worker works pretty much like normal functions now.


src/app/worker-access.ts
```diff
@@ -0,0 +1,16 @@
+import { wrap } from "comlink";
+import type { WorkerWrapper } from "@common/worker-wrapper";
+
+export class WorkerAccess {
+  private static worker: WorkerWrapper | undefined;
```
jienius: Singleton so only one worker at a time

test/worker/test-worker.ts
```diff
@@ -0,0 +1,41 @@
+import test from "ava";
+import * as sinon from "sinon";
+import { WorkerWrapperImpl } from "@worker/worker";
+import Storage from "@worker/storage";
+import { StoredWordList } from "@worker/storage";
+import * as compiler from "@predictive-text-studio/lexical-model-compiler";
+
```
jienius, line 7: I tried to mock storage and test saveFile as well but File doesn't exist on node xD

src/common/worker-wrapper.ts
```diff
@@ -0,0 +1,13 @@
...
+export interface WorkerWrapper {
+  /**
+   * Save a file into IndexedDB
+   */
+  saveFile(name: string, file: File): Promise<ArrayBuffer>;
+  /**
+   * Compile the lexical model using files in the IndexedDB
+   */
+  compileModel(): Promise<string>;
+}
```
eddieantonio: I REALLY like exposing the worker's interface this way! 👍 It's very clean!

Though, now I think we're bumping up against [the second most difficult problem in computer science](https://martinfowler.com/bliki/TwoHardThings.html). 

How about `PredictiveTextStudioWorker`?

```suggestion
export interface PredictiveTextStudioWorker {
  /**
   * Save a file into IndexedDB
   */
  saveFile(name: string, file: File): Promise<ArrayBuffer>;
  /**
   * Compile the lexical model using files in the IndexedDB
   */
  compileModel(): Promise<string>;
}
```

That way the app code uses the `Remote<PredictiveTextStudioWorker>`

src/app/worker-access.ts
```diff
@@ -0,0 +1,16 @@
...
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  private constructor() {}
+
+  static getWorker(): WorkerWrapper {
+    if (!this.worker) {
+      this.worker = wrap(new Worker("worker.js"));
+    }
+    return this.worker;
+  }
+}
```
eddieantonio: Honestly, singleton pattern JavaScript/TypeScript is so uncommon! I think I would prefer just a function:

```suggestion
export function spawnNewWorker(): Remote<WorkerWrapper> {
  return wrap(new Worker("worker.js"))
}
```

Create the wrapped worker instance in `src/app/main.ts` and then explicitly _pass_ in an instance of a `Remote<WorkerWrapper>` to everything that needs it.

---

I have a rant about Singleton patterns, but I'll save that for a later time if you want to hear it lol. If we want to _ensure_ that there's only one Worker ever instantiated by the application, there's a better way to do it in JavaScript. But I think for now, I think even _that_ is overkill.

src/worker/worker.ts
```diff
@@ -0,0 +1,37 @@
...
+  compileModelFromLexicalModelSource,
+  WordListFromArray,
+} from "@predictive-text-studio/lexical-model-compiler";
+import { linkStorageToKmp } from "./link-storage-to-kmp";
+import Storage from "./storage";
+
+export class WorkerWrapperImpl implements WorkerWrapper {
+  private storage: Storage;
+
+  constructor(storage = new Storage()) {
```
eddieantonio: Dependency injection, niiice! 🤩

src/worker/worker.ts
```diff
@@ -0,0 +1,37 @@
...
+    this.storage = storage;
+  }
+
+  saveFile = async (name: string, file: File): Promise<ArrayBuffer> => {
```
eddieantonio: Why use this syntax? Not everything has to be an arrow function 😂 

```suggestion
  async saveFile(name: string, file: File): Promise<ArrayBuffer> {
```

src/worker/worker.ts
```diff
@@ -0,0 +1,37 @@
...
+    const wordlist = await readExcel(await file.arrayBuffer());
+    this.storage.saveFile(name, wordlist);
+    return await linkStorageToKmp(this.storage);
+  };
+
+  compileModel = async (): Promise<string> => {
```
eddieantonio: Same thing here. I'm very confused about the arrow function syntax being used here!

```suggestion
  async compileModel(): Promise<string> {
```

src/common/worker-wrapper.ts
eddieantonio: I have... issues with our current interface... but I think these issues are beyond the scope of this PR 😅

src/worker/worker.ts
```diff
@@ -0,0 +1,37 @@
...
+    if (storedFiles.length < 1) {
+      throw new Error("Cannot find any file in the IndexedDB");
+    } else {
+      const file = storedFiles[0];
+      const code = compileModelFromLexicalModelSource({
+        format: "trie-1.0",
+        sources: [new WordListFromArray(file.name, file.wordlist)],
+      });
+      return code;
+    }
```
eddieantonio: After reading this, I realized that the two functions we export as part of the worker are _wacky_ and make no sense together, but, as I mentioned before, I think that's an issue for a different PR...

test/worker/test-worker.ts
eddieantonio: Yeah :/

test/worker/test-worker.ts
```diff
@@ -0,0 +1,41 @@
...
+test("compile model should throws error when no file is found in the IndexedDB", async (t) => {
+  const storageStub = new Storage();
+  sinon.stub(storageStub, "fetchAllFiles").returns(Promise.resolve([]));
+
+  const workerWrapper = new WorkerWrapperImpl(storageStub);
+
+  const error = await t.throwsAsync(workerWrapper.compileModel());
+  t.is(error.message, "Cannot find any file in the IndexedDB");
```
eddieantonio, line 15: I'm hesitant to test the _exact_ string of an error message like this, but... maybe it doesn't matter here because the error message is unlikely to change? Maybe?

test/worker/test-worker.ts
```diff
@@ -0,0 +1,41 @@
...
+  const storageStub = new Storage();
+  sinon
+    .stub(storageStub, "fetchAllFiles")
+    .returns(Promise.resolve([testStoredFile]));
+  const compilerMock = sinon.mock(compiler);
+  compilerMock
+    .expects("WordListFromArray")
+    .once()
+    .withArgs(testStoredFile.name, testStoredFile.wordlist);
+  compilerMock.expects("compileModelFromLexicalModelSource").once();
```
eddieantonio, line 34: Honestly, I'm not sure what this is exactly trying to exercise. I would test behaviour rather than, did this call this thing I know it will call. 

I think I'd prefer assertions that treat the function as a black box and verify properties about the input and the expected output. Look at `test-storage` to see how a geniune `Storage` class is instantiated and used to test the underlying functionality.

eddieantonio (approved):
L(O)TM (Looks okay to me).

So I will approve because this genuinely fixes #74; however, I can nitpick the specifics all day! Instead of holding up this PR for another week, I say merge it, and rework/refactor it in future PRs. I _really_ like the `WorkerWrapper` interface, but reading it made me realize that the current ways of interacting with the worker is wonky af. I think that's best handled in new issue(s). Also, the Gang of Four Singleton pattern is just bizarre in JavaScript/TypeScript, especially implemented the way it's implemented here. 

But alas, I can rant on, but the worker/app interface has been established here, and in a rather nice way! Good work finding that `comlink` library!

src/app/worker-access.ts
jienius: I remember having this conversation before...

PR #95, opened 2020-10-03
Create conventions document

### LINKED ISSUE #93: Create a general styleguide
We have many unwritten convetions, and it'd be best to document them all!

We need a document, say `docs/conventions.md` that lists these unwritten convetions. Things that a linter _can't_ catch. For example, do we write it `URL` or `Url` in an identifier? How do we name our CSS selectors? How do we order rules in a CSS ruleset? When do we use arrow functions vs. function declarations? etc.

eddieantonio:
Fixes #93

### DIFF
--- a/docs/conventions.md
+++ b/docs/conventions.md
@@ -26,11 +26,99 @@ function fetchJSON() {
 ```
 
 
+General: Always delete commented-out code
+-----------------------------------------
+
+Commented-out code should never be committed to `production`. If there
+is some reason for keeping around the code that was commented out, **use
+git history**
+
+### How to delete code, but keep it in history
+
+If you need to comment out some code, make sure your working copy is
+clean.
+
+    $ git status
+    On branch production
+    Your branch is up to date with 'origin/production'.
+
+    nothing to commit, working tree clean
+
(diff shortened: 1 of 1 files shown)


docs/conventions.md
```diff
@@ -0,0 +1,290 @@
...
+
+When it appears in the middle of an identifier, **only capitalize the first letter**:
+
+```javascript
+// good
+function fetchJson() {
+}
+// bad
+function fetchJSON() {
+}
```
eddieantonio: Looking at this convention, I actually quite dislike it! I'd like to change the recommendation to the OPPOSITE of this; that is, we always capitalize acronyms and initialisms, including:

```javascript
/* good */
function XMLHTTPRequest() {
}
/* bad */
function XmlHttpRequest() {
}
```

I just like the way the caps look better ¯\\\_(ツ)\_/¯

jienius: I like this

jienius (approved):
LGTM

eddieantonio:
In our slack, we have a 3-1 vote on `fetchJSON()` vs `fetchJson()` (1 abstained). So I reversed the decision in the doc! Merging...

PR #96, opened 2020-10-06
add param modelId

### LINKED ISSUE #91: generateKmpJson() should take in modelID as a parameter
Follow up from discussion here: https://github.com/eddieantonio/predictive-text-studio/pull/78#discussion_r498891567

Basically, `modelID` can be generated outside of `generateKmpJson()`, but this function decides to automatically generate a `modelID`, regardless of the input. Instead, it should only generate a `modelID` if none has been explicitly provided.

xqiu1:
Fixes #91


src/worker/generate-kmp.ts
```diff
@@ -24,9 +24,7 @@ export async function generateKmp(
 ): Promise<ArrayBuffer> {
   const kmpJsonFile = generateKmpJson({
     languages: [{ name: langName, id: bcp47Tag }],
-    // TODO: change generateKmpJson() to accept a modelID argument instead of splitting it here:
-    authorID: modelId.split(".")[0],
-    modelTag: modelId.split(".")[2],
+    modelId: modelId,
```
jienius: Might want to make it ID to keep it consistent

xqiu1: Have we decided to use which pattern? modelId or modelID? I found in different files we use different patterns 😂  @eddieantonio

eddieantonio: After the informal Slack poll, I decided we're sticking to `modelID`, even though we used `modelId` in the past. This is now part of [the new conventions document](https://github.com/eddieantonio/predictive-text-studio/blob/production/docs/conventions.md#general-keep-capitals-intact-in-names-with-acronyms-and-initialisms-in-camelcase)! 🥳

src/common/kmp-json-file.ts
```diff
@@ -37,6 +37,11 @@ export interface RelevantKmpOptions {
    * The model version. The first released version **MUST** be 1.0.0.
    */
   version?: string;
+  /**
+   * The model Id.
+   * keyman will create a model Id for every models, derived from "organization, name of language, modelname"
+   */
+  modelId?: string;
```
eddieantonio:
```suggestion
  modelID?: string;
```

src/worker/generate-kmp-json.ts
```diff
@@ -64,7 +64,8 @@ export function generateKmpJson(options: Readonly<RelevantKmpOptions>): string {
   const modelTag = options.modelTag || primaryLanguageISO639;
 
   const authorID = options.authorID || generateAuthorID(options.authorName);
-  const modelID = `${authorID}.${primaryLanguageISO639}.${modelTag}`;
+  const modelID =
+    options.modelId || `${authorID}.${primaryLanguageISO639}.${modelTag}`;
```
eddieantonio:
```suggestion
  const modelID =
    options.modelID || `${authorID}.${primaryLanguageISO639}.${modelTag}`;
```

src/worker/generate-kmp.ts
eddieantonio:
```suggestion
    modelID: modelId,
```

The existing variable, called `modelId` can be changed in a different pull request!

eddieantonio (changes requested):
I know the rules literally flipped over-night, but please switch all new names to `modelID` instead of `modelId`; you do not have to change existing variable names.

Another thing I'd like to see in this PR is a test case added to  [test-generate-kmp.ts](https://github.com/eddieantonio/predictive-text-studio/blob/production/test/test-generate-kmp.ts). The `modelID` parameter determines the _filename_ of the model.js file in the `.kmp` archive. We should have an assertion that, if a custom `modelID` (that differs from what would automatically be generated by `generateKmpFile()`) actually appears the generated filename!

eddieantonio:
Hey @Xuechunqiu! Could you please change `modelId` -> `modeID` in the remaining places? **Tip**: if you go to the "Files changed" tab, you can **batch** all suggestions and make one commit! https://github.com/eddieantonio/predictive-text-studio/pull/96/files

Xuechunqiu:
> Hey @Xuechunqiu! Could you please change `modelId` -> `modeID` in the remaining places? **Tip**: if you go to the "Files changed" tab, you can **batch** all suggestions and make one commit! https://github.com/eddieantonio/predictive-text-studio/pull/96/files

eddieantonio (approved):
LGTM!

src/common/kmp-json-file.ts
```diff
@@ -37,6 +37,11 @@ export interface RelevantKmpOptions {
    * The model version. The first released version **MUST** be 1.0.0.
    */
   version?: string;
+  /**
+   * The model Id.
+   * keyman will create a model Id for every models, derived from "organization, name of language, modelname"
```
eddieantonio: This can be updated in a new PR, but this you might want to refer to this documentation here: https://help.keyman.com/developer/current-version/guides/lexical-models/tutorial/step-2#toc--the-model-id-

test/test-generate-kmp.ts
```diff
@@ -26,4 +26,11 @@ test("it should generate a kmp file", async (t) => {
 
   t.assert(isKmpExist != null);
   t.assert(isModelExist != null);
+
+  //The modelID parameter determines the filename of the model.js file in the .kmp archive
+  if (isKmpExist) {
+    isKmpExist.async("string").then(function (data) {
+      t.assert(data.includes(modelID) == true);
+    });
+  }
```
eddieantonio: This test does _not_ exercise the funtionality implemented.

The functionality implemented affects the _filename_ of the `kmp.json` file in the zip file, however, this test asserts whether the string is contained somewhere within the `kmp.json`.

There are two steps that I'd like to see changed:

 1. revert 7a99a905c86f412acc20ffea06cd24aa6754de94
 2. create a _new_ test case (do not modify the existing one, as is done here), that calls `generateKmp()` with a _different_ modelID that what would normally be generated. Be creative here! Regardless of what the modelID is, the test should assert that the given modelID is contained in the filename of ths .js file.
 3. for extra credit, you could parse the `kmp.json` and look for the given modelID in the `files[i].name` path, where `i` is some index into the array; however, this step is optional.

eddieantonio (changes requested):
I think the test case must be improved. Feel free to ping me on Slack if you have questions!

src/common/kmp-json-file.ts
```diff
@@ -32,11 +31,14 @@ export interface RelevantKmpOptions {
    * The model's copyright string.
    */
   copyright?: string;
-
   /**
    * The model version. The first released version **MUST** be 1.0.0.
    */
   version?: string;
+  /**
+   * The model Id.
```
eddieantonio, line 39: What _is_ the model ID? What does the model ID mean?

test/test-generate-kmp.ts
```diff
@@ -13,17 +13,37 @@ const sources = [
...
 test("it should generate a kmp file", async (t) => {
-  const kmp = await generateKmp(langName, bcp47Tag, sources, modelId);
+  const kmp = await generateKmp(langName, bcp47Tag, sources, modelID);
   t.assert(kmp.byteLength > 0);
 
   const new_zip = new JSZip();
   const zip = await new_zip.loadAsync(kmp);
   const isKmpExist = zip.file("kmp.json");
-  const isModelExist = zip.file("nrc.en.mtnt.model.js");
+  const isModelExist = zip.file(modelID + ".model.js");
```
eddieantonio:
```suggestion
  const isModelExist = zip.file(`${modelID}.model.js`);
```

test/test-generate-kmp.ts
```diff
@@ -13,17 +13,37 @@ const sources = [
...
   t.assert(isKmpExist != null);
   t.assert(isModelExist != null);
 });
+
+const modelID2 = "national_research_council_canada.str.sencoten";
+
+test("The modelID should determines the filename of the model.js file in kmp archive", async (t) => {
+  const kmp = await generateKmp(langName, bcp47Tag, sources, modelID2);
+  const new_zip = new JSZip();
+  const zip = await new_zip.loadAsync(kmp);
```
eddieantonio: I know this was present in the last test, but why is `new_zip` in `snake_case` and not in `camelCase` like all other variable names?

test/test-generate-kmp.ts
```diff
@@ -13,17 +13,37 @@ const sources = [
...
+  const kmpFile = zip.file("kmp.json");
+  const modelFile = zip.file(modelID2 + ".model.js");
```
eddieantonio:
```suggestion
  const modelFile = zip.file(`${modelID2}.model.js`);
```

test/test-generate-kmp.ts
```diff
@@ -13,17 +13,37 @@ const sources = [
...
+
+  t.assert(kmpFile != null);
+  t.assert(modelFile != null);
+
+  if (kmpFile) {
+    await kmpFile.async("string").then(function success(data) {
+      const content = JSON.parse(data);
```
eddieantonio: Why does this use _both_ `await` and a `.then()` Promise callback? If you're going to use the `async`/`await` syntax, I think you should use it consistently within the same function/test case, unless there's a good reason to mix both.

```suggestion
    const data = await kmpFile.async("string")
    const content = JSON.parse(data);
```

test/test-generate-kmp.ts
```diff
@@ -13,17 +13,37 @@ const sources = [
...
+  const kmp = await generateKmp(langName, bcp47Tag, sources, modelID2);
+  const new_zip = new JSZip();
+  const zip = await new_zip.loadAsync(kmp);
+  const kmpFile = zip.file("kmp.json");
+  const modelFile = zip.file(modelID2 + ".model.js");
+
+  t.assert(kmpFile != null);
+  t.assert(modelFile != null);
+
+  if (kmpFile) {
```
eddieantonio, line 43: I think it's better to fail the test here if `kmpFile` is false-y:

```suggestion
  if (!kmpFile) {
    t.fail("could not load kmp.json from KMP archive");
  }
```

**NOTE**: You'd have to adjust the rest of the test case syntax if you accept this suggestion!

eddieantonio (approved):
I think there are some nitpicky details that could be fixed, but other than that, LGTM!

eddieantonio (approved):
LGTM!

Issue #98, opened 2020-10-08
Extract some global styles


eddieantonio:
There are a few common things that can be pulled out into `global.css` such as link styles and utility classes.

eddieantonio:
Resolved by #131, #101, and #99.

Issue #97, opened 2020-10-08
Create functional landing page


eddieantonio:
Instead of the placeholder we currently have, we should have the actual landing page!

See: https://www.figma.com/file/0ifTqGA2I3Y216eCJxh1rr/Keyman-Predictive-Text?node-id=1%3A2

<img width="1540" alt="Figma mockup of the landing page" src="https://user-images.githubusercontent.com/2294397/95503616-cfa6dc80-0968-11eb-9d31-f9c4e4612592.png">

PR #99, opened 2020-10-08
change(app): clean up landing page

eddieantonio:
Removes some old, placeholder stuff, like our names, and the test case for the file input that doesn't actually do anything.

The changes to the landing page markup prepare it for #97, and some of the changes here will help me implement #98.


src/app/pages/LandingPage.svelte
```diff
@@ -1,22 +1,46 @@
...
-  <p>Hi, my name is Xuechun!</p>
-  <p>Hi my name is Eddie!</p>
-  <p>Pranav</p>
+      <button type="submit"> Upload </button>
+    </form>
+  </section>
 
-  <input type="file" data-testid="input-file" bind:files />
+  <footer>
+    <p><small> © 2020 <a href="https://github.com/eddieantonio/predictive-text-studio/graphs/contributors">
```
IanSeng, line 42: **[Question]**: I wonder if using decimal code ` &#169;` would make a difference  ?

IanSeng (approved):
👍

src/app/pages/LandingPage.svelte
eddieantonio: As long as the everything that handles this file corrently deals with UTF-8, there should not be any difference. In the case of U+00A9 COPYRIGHT SIGN, I would prefer the `&copy;` HTML/XML entity, as it's more readable than a random decimal code!

PR #100, opened 2020-10-08
change(app): page routes

eddieantonio:
(work in progress; I'll edit the description when I think this is ready!)


eddieantonio:
I'm not pleased with how this one turned out, so I'm closing it :/

PR #101, opened 2020-10-09
refactor(app): extract .button styles

eddieantonio:
Extracts most of the `Button` component's styles to regular ol' CSS. This means these styles can be reused in other places, like the `<a>` on the landing page.

I used BEM conventions to create the `button` BLOCK, with a number of MODIFIERS, as authored by @pranavnarang. There are no ELEMENTs to the `button` BLOCK, however the `Button` component still exists, and has a "subtext" attribute that acts like an element of the button.

**EDIT**: I also snuck in a fix to the sidebar. For some reason that eludes me, it wasn't getting its width assigned from `width: 75px`, but `min-width` works ¯\\\_(ツ)\_/¯  

Addresses #98

### DIFF
--- a/public/global.css
+++ b/public/global.css
@@ -42,21 +42,27 @@ pre, code, samp {
  * Can be applied to either a <button> or <a> element.
  */
 .button {
+  display: inline-block;
   margin: 0 .625em;
   padding: .625rem 1.25rem; /* button--medium size */
   border: 1px solid transparent;
   border-radius: .625rem;
 
-  font-family: var(--main-font), sans-serif;
-  font-size: 1em;
-  font-weight: bold;
+  font: bold 1em var(--main-font), sans-serif;
   text-transform: uppercase;
+  text-decoration: none;
 
   color: var(--default-text-color);
   background-color: var(--lite-white); /* button--grey */
 
   transition-duration: 0.2s;
   cursor: pointer;
+}
+
+/* make <a> .button have a focus ring like <button> elements. */
+a.button:focus {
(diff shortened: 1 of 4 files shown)


src/app/components/Button.svelte
```diff
@@ -95,9 +27,9 @@
 
 <div class="button-layout">
   <button
-    class="{color} {size}"
-    class:shadow={hasDropShadow}
-    class:outline={isOutlined}
+    class="button button--{color} button--{size}"
```
IanSeng, line 30: love this implementation

public/global.css
```diff
@@ -32,3 +35,75 @@ h1, h2, h3, h4, h5, h6 {
...
+ */
+.button {
+  display: inline-block;
+  margin: 0 .625em;
+  padding: .625rem 1.25rem; /* button--medium size */
+  border: 1px solid transparent;
+  border-radius: .625rem;
+
+  font: bold 1em var(--main-font), sans-serif;
+  text-transform: uppercase;
```
IanSeng, line 52: **[Question]**: is all button text in uppercase?

IanSeng (approved):
just a question about button text in uppercase, everything else LGTM.

public/global.css
eddieantonio: Yes. This is as it was in @pranavnarang's [original design](https://github.com/eddieantonio/predictive-text-studio/blob/2709b17ccd8c77be63212028a0fc95c83b437396/src/app/components/Button.svelte#L14-L22). I haven't given it much thought ¯\\\_(ツ)\_/¯

PR #103, opened 2020-10-10
Factor out upload

### LINKED ISSUE #87: The logic in Upload.svelte needs to be factored out
`Upload.svelte` has become a god-component. In particular, its `<script>` section has become much to large, with many functions that are better extracted and put into their own `.ts` file. After extraction, some of the duplicated logic here can itself be refactored into functions, etc.

---

Refactoring the logic _out_ of `Upload.svelte` will be helpful as most of the functionality of this component will be replaced by components created in #58.

xqiu1:
Fixes #87


pranavnarang (approved):
LGTM!

jienius (approved):
I am not sure about this, but I think we might be able to take advantage of [Svelte store]( https://svelte.dev/tutorial/writable-stores)

Instead of passing in URL into the KMP button from the Upload component, we can create the URL inside the KMP button component itself. When a new file is uploaded, we update the store from Upload and cause the KMP button to generate a new URL. Or we can even generate the KMP file only after the button is clicked. This way, the 2 component is less coupled. 

This doesn't have to be done in this PR, just something we might consider here. @eddieantonio what do you think?

Other than that, LGTM

eddieantonio:
> I am not sure about this, but I think we might be able to take advantage of [Svelte store](https://svelte.dev/tutorial/writable-stores)
> 
> Instead of passing in URL into the KMP button from the Upload component, we can create the URL inside the KMP button component itself. When a new file is uploaded, we update the store from Upload and cause the KMP button to generate a new URL. Or we can even generate the KMP file only after the button is clicked. This way, the 2 component is less coupled.
> 
> This doesn't have to be done in this PR, just something we might consider here. @eddieantonio what do you think?

Yes, we can certainly do something based on Svelte stores! However, this creates a coupling on stores! Haha, we can discuss the various ways to reorganize this in a future PR 😅

PR #114, opened 2020-10-18
Set kmp lang from bcp47

### LINKED ISSUE #111: Set the KMP's language to the entry from the BCP-47 tag on the landing page
I don't think we end up doing anything with the BCP-47 tag input on the landing page. However tag this should be used when generating the KMP file!

Note, this might interact with #109 and #108.

xqiu1:
Fixes #111


src/worker/storage.ts
```diff
@@ -2,7 +2,7 @@ import Dexie, { DexieOptions } from "dexie";
 import { WordList } from "@common/types";
 
 const DB_NAME = "dictionary_sources";
-const SCHEMA_VERSION = 1;
+const SCHEMA_VERSION = 2;
```
jienius, line 5: I think we can get away with this by just deleting the Indexed DB in the browser

xqiu1: you mean I can keep the version = 1?

jienius: right, since there are no actual users at the moment

xqiu1: But when I changed the version number to 1 and downloaded the kmp file, bcp47Tag didn't update, it still used the initial value "".

jienius: You need to delete the old indexed DB, I can get the new tag after I have done that
```
[{"name":"English dictionary","id":"unknownauthor.aa.example","languages":[{"name":"English","id":"aa"}]}]}
```

xqiu1: You're right!

src/app/components/BCP47Tag.svelte
```diff
@@ -1,11 +1,19 @@
...
+  import worker from "../spawn-worker";
 
   const TAG_INPUT_ID = "tag-input";
 
   let bcp47Tag = "";
+  let invalidTag = true;
   $: schema = bcp47Parser.parse(bcp47Tag);
   $: invalidTag = bcp47Tag !== "" && schema.language === null;
+  $: {
+    if (!invalidTag) {
```
eddieantonio: I almost mentioned this during our pair session, but could you **flip* this boolean variable? Instead of

```
invalidTag = true; // the tag is not valid
```

It's simpler to think of:

```
validTag = false
```

(this suggestion fixes a bug where an empty tag was considered a valid tag):

```suggestion
  let validTag = false;
  $: schema = bcp47Parser.parse(bcp47Tag);
  $: validTag = bcp47Tag !== "" && schema.language != null;
  $: {
    if (validTag) {
```

I believe there's code later on that makes use of `invalidTag`, so you'll have to update that too!

src/app/components/BCP47Tag.svelte
```diff
@@ -1,11 +1,19 @@
...
+      console.log("has valid tag in bcp47Tag", bcp47Tag);
```
eddieantonio: Please avoid letting `console.log()` used for debug from going into production!
```suggestion
```

src/worker/link-storage-to-kmp.ts
```diff
@@ -8,10 +8,15 @@ export async function linkStorageToKmp(storage: Storage): Promise<ArrayBuffer> {
     storage
   );
 
+  const maybePackageInfo = await storage.fetchPackageInfo();
+  if (maybePackageInfo == undefined) {
+    throw new Error("the packageInfo is undefined");
```
eddieantonio: I know we wrote this together, but a better error message would indicate what exactly is wrong, and why it might be wrong:

```suggestion
    throw new Error("could not fetch the package info; did you update the package BCP-47 tag?");
```

src/worker/storage.ts
```diff
@@ -71,4 +90,19 @@ export default class Storage {
   fetchAllFiles(): Promise<StoredWordList[]> {
     return this.db.files.toArray();
   }
+  /**
+   * Update bcp47 tag to database
+   * @param bcp47Tag
+   */
+  updateBCP47Tag(bcp47Tag: string): Promise<void> {
+    return this.db.transaction("readwrite", this.db.packageInfo, async () => {
+      await this.db.packageInfo.put({ bcp47Tag, id: 0 });
```
eddieantonio: Could you please extact `0` as a constant?

```typescript
/**
 * The key of the ONLY StoredPackageInfo record.
 */
const PACKAGE_ID = 0;
```

And change this line to:
```suggestion
      await this.db.packageInfo.put({ bcp47Tag, id: PACKAGE_ID });
```

test/test-link-storage-to-kmp.ts
```diff
@@ -10,6 +10,8 @@ test("a function link the storage file to generate the kmp package", async (t) =
     IDBKeyRange,
   });
   const storage = new Storage(db);
+  // make sure the packageInfo is not undefined
```
eddieantonio:
```suggestion
  // make sure the packageInfo defined
```

test/test-storage.ts
```diff
@@ -88,3 +88,24 @@ test("retrieving mulitple files with .fetchAllFiles()", async (t) => {
     return a.name > b.name ? 1 : -1;
   }
 });
+
+test("update the bcp47 tag to database", async (t) => {
+  const { db, storage } = t.context;
+  // At first, nothing in the DB
+  t.is(await db.packageInfo.count(), 0);
+  await storage.updateBCP47Tag("en");
+  // Now there's one file in the DB!
```
```suggestion
  // Now there's one package info record in the DB!
```

test/test-storage.ts
```diff
@@ -88,3 +88,24 @@ test("retrieving mulitple files with .fetchAllFiles()", async (t) => {
...
+test("retrieve bcp47tag from the database", async (t) => {
+  const { storage } = t.context;
+  await storage.updateBCP47Tag("en");
+  const maybePackageInfo = await storage.fetchPackageInfo();
+  if (maybePackageInfo != undefined) {
+    const bcp47Tag = maybePackageInfo.bcp47Tag;
+    t.is(bcp47Tag, "en");
+  } else {
+    t.fail("packageInfo is undefined.");
+  }
```
eddieantonio: Consider a [guard class refactoring](https://wiki.c2.com/?GuardClause):

```suggestion
  if (maybePackageInfo == undefined) {
      t.fail("packageInfo is undefined.");
  }

  const bcp47Tag = maybePackageInfo.bcp47Tag;
  t.is(bcp47Tag, "en");
```

src/worker/storage.ts
```diff
@@ -71,4 +90,19 @@ export default class Storage {
...
+    });
+  }
+  /**
+   * Retrieves packageInfo in the database
+   */
+  fetchPackageInfo(): Promise<StoredPackageInfo | undefined> {
+    return this.db.packageInfo.where(":id").equals(0).first();
```
```suggestion
    return this.db.packageInfo.where(":id").equals(PACKAGE_ID).first();
```

src/worker/storage.ts
```diff
@@ -42,8 +51,18 @@ export class PredictiveTextStudioDexie extends Dexie {
        * +------------------+
        */
       files: "++id, name, wordlist",
+      /**
+       * packageInfo Table Scehma
```
eddieantonio, line 59: What does `packageInfo` store? What is the intended purpose of what we store here? What might go in here in the future?

eddieantonio (approved):
L(okay)TM!

There's a few refactorings and changes I suggest to make things neater 😌 which I _do_ expect you to implement or ask me questions on, but after you've incorporated these changes, feel free to merge :)

src/worker/storage.ts
eddieantonio: I'd like to get in the habit of incrementing the version ID every time we make a change to the schema, because "just delete your indexedDB" is not always going to be an option!

test/test-storage.ts
```diff
@@ -88,3 +88,24 @@ test("retrieving mulitple files with .fetchAllFiles()", async (t) => {
...
+test("retrieve bcp47tag from the database", async (t) => {
+  const { storage } = t.context;
+  await storage.updateBCP47Tag("en");
+  const maybePackageInfo = await storage.fetchPackageInfo();
+  if (maybePackageInfo == undefined) {
+    t.fail("packageInfo is undefined.");
+  } else {
+    const bcp47Tag = maybePackageInfo.bcp47Tag;
+    t.is(bcp47Tag, "en");
+  }
```
xqiu1, line 110: If maybePackageInfo is not surrounded by else{}, an error will be reported.

eddieantonio: Ah, my bad! I thought `t.fail()` had [the `never` type](https://www.typescriptlang.org/docs/handbook/basic-types.html#never), but [it does not](https://github.com/avajs/ava/blob/b71f396df7b9d74ef24262b6840df1593cad3798/index.d.ts#L141). Kinda annoying :/

The if/else will have to do!

src/app/components/BCP47Tag.svelte
```diff
@@ -1,11 +1,18 @@
...
   import bcp47Parser from "bcp-47";
+  import worker from "../spawn-worker";
 
   const TAG_INPUT_ID = "tag-input";
 
   let bcp47Tag = "";
+  let validTag = false;
   $: schema = bcp47Parser.parse(bcp47Tag);
-  $: invalidTag = bcp47Tag !== "" && schema.language === null;
+  $: validTag = bcp47Tag !== "" && schema.language === null;
```
eddieantonio: The BCP-47 tag is valid when it is has content (not empty) and it has a parsed `language`:

```suggestion
  $: validTag = bcp47Tag.length > 0 && schema.language != null;
```

src/worker/storage.ts
```diff
@@ -42,8 +51,20 @@ export class PredictiveTextStudioDexie extends Dexie {
...
+       * packageInfor stores the information needed for create the kmp package.
+       * we might also store language name, moded id in the future.
```
```suggestion
       * packageInfo stores the information needed for creating  the KMP package.
       * we might also store language name, model ID in the future.
```

src/worker/storage.ts
```diff
@@ -71,4 +92,27 @@ export default class Storage {
...
+    });
+  }
+  /**
+   * Retrieves packageInfo in the database
+   */
+  fetchPackageInfo(): Promise<StoredPackageInfo | undefined> {
+    /**
+     * The key of the ONLY StoredPackageInfo record.
+     */
+    const PACKAGE_ID = 0;
```
eddieantonio: Since the `PACKAGE_ID` constant represents the exact same value across two different methods, it should be extracted to the module scope to DRY things up:

```suggestion
  /**
   * Update bcp47 tag to database
   * @param bcp47Tag
   */
  updateBCP47Tag(bcp47Tag: string): Promise<void> {
    return this.db.transaction("readwrite", this.db.packageInfo, async () => {
      await this.db.packageInfo.put({ bcp47Tag, id: PACKAGE_ID });
    });
  }
  /**
   * Retrieves packageInfo in the database
   */
  fetchPackageInfo(): Promise<StoredPackageInfo | undefined> {
```

And then place this _above_ the `class` definition:

```typescript
/**
 * The key of the ONLY StoredPackageInfo record.
 */
const PACKAGE_ID = 0;
```

eddieantonio (changes requested):
Please address the remaining issues, and then we can merge!

src/app/components/BCP47Tag.svelte
```diff
@@ -1,11 +1,20 @@
 <script lang="ts">
   import bcp47Parser from "bcp-47";
+  import worker from "../spawn-worker";
 
   const TAG_INPUT_ID = "tag-input";
 
   let bcp47Tag = "";
+  let validTag = false;
+  let displayEnterBCP47Tag = true;
```
xqiu1, line 9: I added this variable to control the displayed message, since I think we should display 'Enter BCP 47 tag' when the input is empty.

eddieantonio: Excellent!

eddieantonio (approved):
LGTM! Merge when ready!

cypress/integration/upload.spec.ts
```diff
@@ -39,7 +39,7 @@ describe("Upload from the the landing page", function () {
 
     cy.get("@download-kmp")
       .should("have.attr", "data-download-state", "ready")
-      .click();
+      .click({ force: true });
```
xqiu1, line 45: The cypress test failed and the error message said cy.click() failed because this element is not visible. I add force:true to ignore it... I don't know why this test failed, it always passed before.

cypress/integration/upload.spec.ts
```diff
@@ -16,6 +16,9 @@ describe("Upload from the the landing page", function () {
 
   it("should find a button to press to upload a file", function () {
     cy.visit("/");
+    // we need a valid bcp47Tag to update the packageInfo
+    cy.contains("label", "Enter BCP 47 tag").scrollIntoView();
+    cy.get("#tag-input").type("zh", { force: true });
```
xqiu1, line 21: Since I changed the logic in BCP47Tag.svelte, the packageInfo will update the bcp47Tag only when a valid tag is provided (tag can not be empty). If we don't enter a valid tag, the packageInfo will be undefined, which causes error when we try to download the KMP file.

PR #118, opened 2020-10-19
fix(worker): await for the file to be saved before compiling the model

eddieantonio:
This fixes a bug I noticed. Normally, I'd write a bug report and a regression test, but since this tests a race condition, I honestly don't know how to consistently reproduce the bug, other than to upload a giant spreadsheet every time :/


IanSeng (approved):
😃

PR #122, opened 2020-10-26
Figure out what methods should exist between app and worker

### LINKED ISSUE #108: Figure out what methods should exist between the app and the worker!
Right now `compileModel()` and `saveFile()` are wonky. Addtionally, there probably should be events that fire and update stores on the front-end.

Please ask me, @eddieantonio, to clarify my thoughts on this issue, but I think it should be addressed sooner rather than later! With tests!

xqiu1:
Fixes #108

### DIFF
--- a/src/app/components/BCP47Tag.svelte
+++ b/src/app/components/BCP47Tag.svelte
@@ -12,7 +12,7 @@
   $: displayEnterBCP47Tag = bcp47Tag == "" || validTag;
   $: {
     if (validTag) {
-      worker.updateBCP47Tag(bcp47Tag);
+      worker.setProjectData({languages: [{ name: "", id: bcp47Tag }]});
     }
   }
 </script>
--- a/src/app/components/BCP47Tag.svelte
+++ b/src/app/components/BCP47Tag.svelte
@@ -12,7 +12,7 @@
   $: displayEnterBCP47Tag = bcp47Tag == "" || validTag;
   $: {
     if (validTag) {
-      worker.setProjectData({languages: [{ name: "", id: bcp47Tag }]});
+      worker.setProjectData({ languages: [{ name: "", id: bcp47Tag }] });
     }
   }
 </script>
--- a/src/common/predictive-text-studio-worker.ts
+++ b/src/common/predictive-text-studio-worker.ts
@@ -19,10 +19,6 @@ export interface PredictiveTextStudioWorker {
    */
   updateBCP47Tag(bcp47Tag: string): Promise<void>;
   /**
-   * Generate a KMP file and return a blob of it
-   */
-  generateKMPFromStorage(): Promise<Blob | void>;
-  /**
    * Take a dictionary source and store it
    * @param name
    * @param contents
@@ -43,7 +39,7 @@ export interface PredictiveTextStudioWorker {
    * Register a callback that is called when the KMP package has succesfully compiled
    * @param callback
    */
-  onPackageCompileSuccess(callback: (kmp: Blob) => void): void;
+  onPackageCompileSuccess(callback: (kmp: ArrayBuffer) => void): void;
   /**
    * Sets optional and required metadata such as BCP-47, language name, author name, copyright string, etc.
    * @param metadata
--- a/src/common/predictive-text-studio-worker.ts
+++ b/src/common/predictive-text-studio-worker.ts
@@ -1,12 +1,15 @@
+import { RelevantKmpOptions } from "@common/kmp-json-file";
 /**
  * This interface establishes a communication protocal between the UI thread and the worker thread
  */
 export interface PredictiveTextStudioWorker {
   /**
+   * @deprecated removed in future
    * Save a file into IndexedDB
    */
   saveFile(name: string, file: File): Promise<ArrayBuffer>;
   /**
+   * @deprecated removed in future
    * Compile the lexical model using files in the IndexedDB
    */
   compileModel(): Promise<string>;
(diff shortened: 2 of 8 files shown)


xqiu1:
@eddieantonio This is just a version to see if I'm on the right track 😅 Are you available to have a look at this?

src/worker/predictive-text-studio-worker-impl.ts
```diff
@@ -39,4 +50,69 @@ export class PredictiveTextStudioWorkerImpl
...
+
+  onPackageCompileError(error: Error, callback: (err: Error) => void): void {
+    // TODO
+    callback(error);
+  }
+
+  onPackageCompileSuccess(kmp: Blob, callback: (kmp: Blob) => void): void {
+    // TODO
+    callback(kmp);
+  }
```
xqiu1: I'm not very sure about what should be done within these methods.

src/common/predictive-text-studio-worker.ts
```diff
@@ -15,4 +18,35 @@ export interface PredictiveTextStudioWorker {
...
   updateBCP47Tag(bcp47Tag: string): Promise<void>;
+  /**
+   * Generate a KMP file and return a blob of it
+   */
+  generateKMPFromStorage(): Promise<Blob | void>;
+  /**
+   * Take a dictionary source and store it
+   * @param name
+   * @param contents
+   * Return how many words were added
```
```suggestion
   * @return how many words were added
```

src/common/predictive-text-studio-worker.ts
```diff
@@ -15,4 +18,35 @@ export interface PredictiveTextStudioWorker {
    * @param bcp47Tag
    */
   updateBCP47Tag(bcp47Tag: string): Promise<void>;
+  /**
+   * Generate a KMP file and return a blob of it
+   */
+  generateKMPFromStorage(): Promise<Blob | void>;
```
eddieantonio: Why was this method added? Who will call it and when?

src/common/predictive-text-studio-worker.ts
```diff
@@ -15,4 +18,35 @@ export interface PredictiveTextStudioWorker {
...
+  /**
+   * Register a callback that is called when the KMP package has succesfully compiled
+   * @param callback
+   */
+  onPackageCompileSuccess(kmp: Blob, callback: (kmp: Blob) => void): void;
+  /**
+   * Sets optional and required metadata such as BCP-47, language name, author name, copyright string, etc.
+   * @param metadata
+   */
+  setProjectData(metadata: Readonly<RelevantKmpOptions>): void;
```
eddieantonio: I like this signature! Though I think it's better to add `Partial` as well:

```suggestion
  setProjectData(metadata: Partial<Readonly<RelevantKmpOptions>>): void;
```

src/worker/predictive-text-studio-worker-impl.ts
```diff
@@ -39,4 +50,69 @@ export class PredictiveTextStudioWorkerImpl
   updateBCP47Tag(bcp47Tag: string): Promise<void> {
     return this.storage.updateBCP47Tag(bcp47Tag);
   }
+
+  async generateKMPFromStorage(): Promise<Blob | void> {
```
eddieantonio: I like this method, but I think it should be `private` and **not** part of the App/Worker interface. 
```suggestion
  private async generateKMPFromStorage(): Promise<Blob | void> {
```

src/worker/predictive-text-studio-worker-impl.ts
eddieantonio: They should *register a callback* — aka, store the callback as a variable/property. These callbacks are called later, e.g., in `generateKMPFromStorage()`

```suggestion
  private _emitPackageCompileStart: () => void = () => {};
  private _emitPackageCompileError: (err: Error) => void = () => {};
  private _emitPackageCompileSuccess: (kmp: Blob) => void = () => {};

  onPackageCompileStart(callback: () => void): void {
    this._emitPackageCompileStart = callback;
  }

  onPackageCompileError(callback: (err: Error) => void): void {
    this._emitPackageCompileError = callback;
  }

  onPackageCompileSuccess(callback: (kmp: Blob) => void): void {
    this._emitPackageCompileSuccess = callback;
  }
```

src/worker/predictive-text-studio-worker-impl.ts
```diff
@@ -39,4 +50,69 @@ export class PredictiveTextStudioWorkerImpl
...
+    // TODO: Parse multiple dictionary sources, right now just reading the first file
+    this.onPackageCompileStart((): void => {
+      console.log("Compile starts.");
+    });
```
eddieantonio: Call the registered callback instead:

```suggestion
    this._emitPackageCompileStart();
```

src/common/predictive-text-studio-worker.ts
```diff
@@ -15,4 +18,35 @@ export interface PredictiveTextStudioWorker {
...
+  /**
+   * Register a callbcak that is called if the KMP package compilation fails
+   * @param callback
+   */
+  onPackageCompileError(error: Error, callback: (err: Error) => void): void;
+  /**
+   * Register a callback that is called when the KMP package has succesfully compiled
+   * @param callback
+   */
+  onPackageCompileSuccess(kmp: Blob, callback: (kmp: Blob) => void): void;
```
```suggestion
  onPackageCompileError(callback: (err: Error) => void): void;
  /**
   * Register a callback that is called when the KMP package has succesfully compiled
   * @param callback
   */
  onPackageCompileSuccess(callback: (kmp: Blob) => void): void;
```

src/worker/predictive-text-studio-worker-impl.ts
```diff
@@ -39,4 +50,69 @@ export class PredictiveTextStudioWorkerImpl
...
+
+    const storedFiles = await this.storage.fetchAllFiles();
+    if (storedFiles.length < 1) {
+      const error = new Error("Cannot find any file in the IndexedDB");
+      this.onPackageCompileError(error, (error: Error): void => {
+        console.log(error);
+      });
```
```suggestion
      this._emitPackageCompileError(new Error("Cannot find any files in the IndexedDB"));
```

src/worker/predictive-text-studio-worker-impl.ts
```diff
@@ -39,4 +50,69 @@ export class PredictiveTextStudioWorkerImpl
...
+    } else {
+      const kmpFile = await linkStorageToKmp(this.storage);
```
eddieantonio, line 69: `linkStorageToKMP()` can raise an error which _should_ be caught and passed to the `this. _emitPackageCompileError` callback.

src/worker/predictive-text-studio-worker-impl.ts
```diff
@@ -39,4 +50,69 @@ export class PredictiveTextStudioWorkerImpl
...
+      const blob = new Blob([kmpFile], { type: "application/octet-stream" });
+      this.onPackageCompileSuccess(blob, (kmp: Blob): void => {
+        console.log("Compile successes.");
+      });
+      return blob;
```
```suggestion
      this._emitPackageCompileSuccess(blob);
```

src/worker/storage.ts
```diff
@@ -28,18 +28,51 @@ export interface StoredWordList {
...
+  /**
+   *
+   */
+  version?: string;
+}
 
 export class PredictiveTextStudioDexie extends Dexie {
   files: Dexie.Table<StoredWordList, number>;
   packageInfo: Dexie.Table<StoredPackageInfo, number>;
+  projectData: Dexie.Table<StoredProjectData, number>;
```
eddieantonio, line 64: Why did you create a new table instead of extending the existing table?

eddieantonio (changes requested):

src/app/components/BCP47Tag.svelte
```diff
@@ -12,7 +12,7 @@
   $: displayEnterBCP47Tag = bcp47Tag == "" || validTag;
   $: {
     if (validTag) {
-      worker.updateBCP47Tag(bcp47Tag);
+      worker.setProjectData({ languages: [{ name: "", id: bcp47Tag }] });
```
xqiu1, line 15: We may use the bcp47Tag to generate the language name, or in an opposite way (let the user to input language name and we generate the bcp47Tag).

IanSeng (approved):

eddieantonio (approved):

Issue #113, opened 2020-10-16
Try migrating unit tests to Mocha or Jest that run IN A BROWSER ENVIRONMENT


eddieantonio:
**EDIT**:

To satisfy this, the tests must be able to access and/or mock the following [`WindowOrWorkerGlobalScope `](https://developer.mozilla.org/en-US/docs/Web/API/Window) properties **without the use of mocks**:

 - [x] [`fetch()`](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch)
 - [ ] [`indexedDB`](WindowOrWorkerGlobalScope.indexedDB)
 - [x] [`new Blob`](https://developer.mozilla.org/en-US/docs/Web/API/Blob/Blob)
 - [ ] [`File.File()`](https://developer.mozilla.org/en-US/docs/Web/API/File/File)

eddieantonio:
Hey @IanSeng, I added `File.File()` to the list, as I forgot it before.

PR #123, opened 2020-10-29
[#36]feat: Fetch a list of existing FirstVoices/Keyman keyboards

### LINKED ISSUE #36: Fetch a list of existing FirstVoices/Keyman keyboards

IanSeng:
# Summary 
- Revisited the branch `feat/keyman-api` to fix feedback to fetch languages data from the Keyman api and store them in IndexedDB by leveraging the package `dexie` instead `idb`
- Moved interfaces from `storage.ts` to `models.ts` to reduce the file size 
- Fetching KMP keyboard data is initialized with the service worker 

### Sub-task 
- [x] Query Keyman api to get languages information 
- [x] Save data into IndexedDB
- [x] Unit testing api 
- [x] Unit testing all function created

### Issue
Resolves #36 & #113


src/worker/keyman-api-service.ts
```diff
@@ -0,0 +1,37 @@
...
+   */
+  params = { q: "l" };
+  url: URL;
+  langageArray: Array<keyboardDataObj>;
+
+  constructor() {
+    this.langageArray = [];
+    this.url = new URL(this.baseUrl);
+    this.url.search = new URLSearchParams(this.params).toString();
+    console.log(this.url.href);
```
pranavnarang: Is this `console.log` needed or was it just used to debug/check?

src/worker/storage.ts
```diff
@@ -111,4 +93,18 @@ export default class Storage {
   fetchPackageInfo(): Promise<StoredPackageInfo | undefined> {
     return this.db.packageInfo.where(":id").equals(PACKAGE_ID).first();
   }
+  /**
+   * Save all keyboard info into the database
+   */
+  addKeybaordData(langauge: string, bcp47: string): Promise<void> {
```
pranavnarang: typo with Keyboard spelling

test/test-worker.ts
```diff
@@ -1,41 +1,60 @@
...
-  t.is(error.message, "Cannot find any file in the IndexedDB");
+test.before("optional title", () => {
+  storageStub = new Storage();
+  keymanApiMock = new KeymanApi();
+  sinon
+    .stub(storageStub, "fetchAllFiles")
+    .returns(Promise.resolve([storedFileStub]));
+  // sinon
+  //   .stub(keymanApiMock, "getLanaguageData")
+  //   .returns(Promise.resolve(keymanKeyboardDataStub));
```
pranavnarang: Can the commented out code be removed?

test/test-worker.ts
```diff
@@ -1,41 +1,60 @@
...
 
   compilerMock.verify();
   t.pass();
 });
+
+// test("fetch and save Keyman keyboard data", async (t) => {
+//   // await keymanApiMock.getLanaguageData().then((aaa) => {
+//   //   console.log(aaa);
+//   // });
+// });
```
pranavnarang: Can the commented out code be removed?

pranavnarang (dismissed):
Can't really comment on the logic as I'm not too sure how the API works yet but it just needs a few other minor changes and should be good to go!

src/worker/models.ts
```diff
@@ -0,0 +1,78 @@
...
+   */
+  copyright?: string;
+
+  /**
+   * The version number of the model.
+   */
+  version?: string;
+}
+
+export interface keyboardDataObj {
```
eddieantonio: What is up with the name of this interface? I guess it's an unwritten rule in the project, but all interfaces should be written in `CapitalCamelCase`. I think this is better as 
```suggestion
export interface KeyboardData {
```

src/worker/keyman-api-service.ts
```diff
@@ -0,0 +1,36 @@
...
+          this.langageArray.push({
+            bcp47Tag: element.id,
+            langauge: element.name,
+          });
+        });
+      })
+      .then(() => {
+        return this.langageArray;
+      });
+  }
```
eddieantonio: Why declare the function `async` if you didn't use `await` in the code?

```suggestion
  async fetchLanaguageData(): Promise<keyboardDataObj[]> {
    let response = await fetch(this.url.href);
    let data: { languages: Array<KeymanObj> } = response.json()
    for (let element of data.languages) {
      this.langageArray.push({
        bcp47Tag: element.id,
        langauge: element.name,
      });
    }

    return this.languageArray;
  }
```

src/worker/keyman-api-service.ts
```diff
@@ -0,0 +1,36 @@
+import { keyboardDataObj, KeymanObj } from "./models";
+
+export class KeymanApi {
```
eddieantonio: Please capitalize "API":

```suggestion
export class KeymanAPI {
```

See: https://github.com/eddieantonio/predictive-text-studio/blob/production/docs/conventions.md#general-keep-capitals-intact-in-names-with-acronyms-and-initialisms-in-camelcase

src/worker/keyman-api-service.ts
```diff
@@ -0,0 +1,36 @@
...
+  /**
+   * API configs
+   */
+  baseUrl = "https://api.keyman.com/search";
+  /**
+   * Query data to searche only for languages that match
```
eddieantonio:
```suggestion
   * Query data to search only for languages that match
```

That match what? Could you please link to their API documentation?

src/worker/models.ts
```diff
@@ -0,0 +1,78 @@
...
+  /**
+   * Keyboard Language of the bcp47Tag
+   */
+  langauge: string;
+}
+
+/**
+ * KMP endpoint returning object type
+ */
+export interface KeymanObj {
```
eddieantonio: Try to base the name off of the documentation. They call this the `SearchLanguage` object. 

```suggestion
export interface SearchLanguage {
```

See: https://help.keyman.com/developer/cloud/search/1.0/#toc-searchlanguage-object

src/worker/models.ts
```diff
@@ -0,0 +1,78 @@
...
+   */
+  bcp47Tag: string;
+  /**
+   * Keyboard Language of the bcp47Tag
+   */
+  langauge: string;
+}
+
+/**
+ * KMP endpoint returning object type
```
eddieantonio: Please link the appropriate documentation here: https://help.keyman.com/developer/cloud/search/1.0/#toc-searchlanguage-object

```suggestion
 * KMP endpoint returning object type
```

src/worker/keyman-api-service.ts
```diff
@@ -0,0 +1,36 @@
...
+   */
+  params = { q: "l" };
+  url: URL;
+  langageArray: Array<keyboardDataObj>;
+
+  constructor() {
+    this.langageArray = [];
```
eddieantonio: Why is this a mutable property of this class? Why does it get written to (via `.push()`) and returned in the same method, but is never emptied or accessed otherwise by the class?

src/worker/predictive-text-studio-worker-impl.ts
```diff
@@ -17,20 +18,32 @@ const defaultCopyright = "";
 function doNothing() {
   // intentionally empty
 }
+import { keyboardDataObj } from "./models";
```
eddieantonio: Weird. I'd expect prettier to hoist this import to the top of the file ¯\\\_(ツ)\_/¯

src/worker/predictive-text-studio-worker-impl.ts
```diff
@@ -17,20 +18,32 @@ const defaultCopyright = "";
...
 
   async updateBCP47Tag(bcp47Tag: string): Promise<void> {
     await this.storage.updateBCP47Tag(bcp47Tag);
     return this.generateKMPFromStorage();
   }
 
+  async getLanguageData(): Promise<void> {
+    await this.keymanApi
+      .fetchLanaguageData()
+      .then((languages: keyboardDataObj[]) => {
```
eddieantonio: Why do you use both `await` and Promise `.then` callbacks that `await` replaces?

src/worker/keyman-api-service.ts
eddieantonio: Could this class be a single function? Classes with one method in addition to their constructor are candidates to be simplified to a simple function.

I can see the use of this class if it allowed for the dependency injection of `fetch()`, but the tests  overwrite `global.fetch` instead, so right now, it seems like a class is unnecessary.

src/worker/models.ts
```diff
@@ -0,0 +1,78 @@
+import { WordList } from "@common/types";
```
eddieantonio: I appreciate this refactor! `storage.ts` was getting a bit too large 😓 

Could you please add a JSDoc comment to the file explaining what's supposed to be stored in here?

src/worker/keyman-api-service.ts
```diff
@@ -0,0 +1,36 @@
...
+  /**
+   * API configs
+   */
+  baseUrl = "https://api.keyman.com/search";
+  /**
+   * Query data to searche only for languages that match
+   */
+  params = { q: "l" };
+  url: URL;
+  langageArray: Array<keyboardDataObj>;
```
eddieantonio: Please use this notation instead:

```suggestion
  langageArray: keyboardDataObj[];
```

Although, I do suggest using a different name for this interface!

test/fixtures/index.ts
```diff
@@ -19,3 +24,32 @@ export const exampleWordlist: WordList = [
...
+
+export const responseBody = {
+  languages: [
+    { id: "km", name: "Khmer (Battambang Khmer)", keyboards: [Array] },
+    {
+      id: "kxm",
+      name: "Khmer, Northern (Khmer Lue)",
+      keyboards: [Array],
+    },
+    { id: "kdt", name: "Kuay (Old Khmer)", keyboards: [Array] },
```
eddieantonio: Please include the actually contents from the response (did you copy/paste from a `console.log`? `console.log` omits information deep within objects :/)

```suggestion
    { id: "km", name: "Khmer (Battambang Khmer)", keyboards:  [
                "basic_kbdkhmr",
                "basic_kbdkni",
                "kbdkhmr",
                "khmer10",
                "khmer_angkor",
                "sil_khmer"
            ]
    },
    {
      id: "kxm",
      name: "Khmer, Northern (Khmer Lue)",
      keyboards: [
                "sil_khmer"
            ],
    },
    { id: "kdt", name: "Kuay (Old Khmer)", keyboards: [
                "sil_khmer"
            ]
     },
```

(this comes from <https://api.keyman.com/search?q=l:Khmer>)

src/worker/keyman-api-service.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export class KeymanApi {
+  /**
+   * API configs
+   */
+  baseUrl = "https://api.keyman.com/search";
+  /**
+   * Query data to searche only for languages that match
+   */
+  params = { q: "l" };
```
eddieantonio: Could you explain how this query will return the results of, say the "Halq'eméylem" keyboard layout for mobile?

test/test-keyman-api-service.ts
```diff
@@ -0,0 +1,29 @@
+import { responseBody } from "./fixtures/index";
```
eddieantonio: Unnecessary to include `index`:

```suggestion
import { responseBody } from "./fixtures";
```

test/test-worker.ts
```diff
@@ -10,8 +13,12 @@ test("it should set project data and update to the database", async (t) => {
     bcp47Tag: "en",
     authorName: "UnknownAuthor",
   } as StoredProjectData;
+  const keymanApiMock = new KeymanApi();
```
eddieantonio: Lol, this isn't a mock — it's the real implementation!

eddieantonio (changes requested):
Needs a bit more refining; let me know if you have any questions!

src/worker/predictive-text-studio-worker-impl.ts
IanSeng: nice catch, only one await is needed here

src/worker/models.ts
IanSeng: not a problem, do you prefer another model file that is specifically for all of the storage models or adding JSDoc this this file is fine?

src/worker/keyman-api-service.ts
IanSeng: this query will return all existing keyboard

src/worker/models.ts
eddieantonio: I like option of having a file for _just_ the models we use in storage! This refactor helps keep the storage models in a cohesive place. I would just like a one line JSDoc at the top of the file like:

```typescript
/**
 * @file Interfaces used in storage
 */
```

src/worker/keyman-api-service.ts
eddieantonio: After reading the documentation, I feel like `?q=l` returns all keyboards whose languages have an "l" in the name. I think `?q=` will return ALL keyboards.

src/worker/keyman-api-service.ts
eddieantonio: I'm still waiting for an explaination for this one. Is it intended that when you call `fetchLanguageData()` more than once, that the languageArray may potentially have many duplicate languages?

src/worker/keyman-api-service.ts
```diff
@@ -0,0 +1,42 @@
...
+  /**
+   * API configs
+   */
+  baseUrl = "https://api.keyman.com/search";
+  /**
+   * Query all languages data, doc: https://help.keyman.com/developer/cloud/search/1.0/
+   */
+  params = { q: "l" };
+  url: URL;
+  langageArray: Array<KeyboardData>;
```
eddieantonio: Typo in the name, plus a better name excludes its type (i.e., an array called "Array"):

```suggestion
  languages: KeyboardData[];
```

src/worker/models.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * @file Interfaces used across the worker files excluding stroage
```
eddieantonio, line 2: 👍🏼

src/worker/models.ts
```diff
@@ -0,0 +1,12 @@
...
+ */
+
+/**
+ * KMP endpoint returning object type. Doc: https://help.keyman.com/developer/cloud/search/1.0/#toc-searchlanguage-object
```
eddieantonio, line 6: 👍🏼

src/worker/predictive-text-studio-worker-impl.ts
```diff
@@ -20,17 +22,26 @@ function doNothing() {
...
     await this.storage.updateBCP47Tag(bcp47Tag);
     return this.generateKMPFromStorage();
   }
 
+  async getLanguageData(): Promise<void> {
+    this.keymanAPI.fetchLanaguageData().then((languages: KeyboardData[]) => {
+      languages.forEach(async (data) => {
+        await this.storage.addKeyboardData(data.langauge, data.bcp47Tag);
+      });
+    });
```
eddieantonio: Every time the worker starts up, it will fetch the entire language list. This is bad because it should really only be fetched maybe once a day, at most. But if you open a new tab, a new worker thread starts, which issues this request to the Keyman API 💀 

Is there a way you can set a limit to how often the data is fetched?

src/worker/storage-models.ts
```diff
@@ -0,0 +1,73 @@
+/**
+ * @file Interfaces used in storage
```
eddieantonio, line 2: I like this refactor!

src/worker/storage.ts
```diff
@@ -111,12 +58,25 @@ export class PredictiveTextStudioDexie extends Dexie {
...
+       * +------------------+
+       * | id (primary key) |
+       * +------------------+
+       * +------------------+
+       * | bcp47Tag         |
+       * +------------------+
+       * | language         |
+       * +------------------+
+       */
+      keyboardData: "++id, bcp47Tag, langauge",
```
eddieantonio: I have a degree in linguistics and my MSc was in language models. I still can't spell "language" right on my first try 💀  — lol so please correct this:

```suggestion
      keyboardData: "++id, bcp47Tag, language",
```

src/worker/storage.ts
```diff
@@ -194,4 +154,20 @@ export default class Storage {
 
     return projectData;
   }
+
+  /**
+   * Save all keyboard info into the database
+   */
+  addKeyboardData(langauge: string, bcp47: string): Promise<void> {
+    return this.db.transaction("readwrite", this.db.keyboardData, async () => {
+      await this.db.keyboardData.put({ langauge, bcp47Tag: bcp47 });
```
```suggestion
      await this.db.keyboardData.put({ language, bcp47Tag: bcp47 });
```

src/worker/storage-models.ts
```diff
@@ -0,0 +1,73 @@
...
+
+export interface KeyboardData {
+  /**
+   * Store bcp47Tag as the primarykey
+   */
+  bcp47Tag: string;
+  /**
+   * Keyboard Language of the bcp47Tag
+   */
+  langauge: string;
```
eddieantonio:
```suggestion
  language: string;
```

test/test-keyman-api-service.ts
```diff
@@ -0,0 +1,29 @@
...
+});
+
+// Sad Path
+test("getLanaguageData to catch the error", async (t) => {
+  nock("https://api.keyman.com")
+    .get("/search?q=l")
+    .replyWithError({ code: "System Error" });
+  await keymanAPI.fetchLanaguageData().catch((response) => {
+    t.is(response.code, "System Error");
+  });
```
eddieantonio: Consider using [`t.throwsAsync()`](https://github.com/avajs/ava/blob/master/docs/03-assertions.md#throwsasyncthrower-expectation-message) instead.

test/test-storage.ts
```diff
@@ -160,3 +160,26 @@ test("retrieve project data from the database", async (t) => {
...
+  t.is(await db.keyboardData.count(), 0);
+  await storage.addKeyboardData(
+    keymanKeyboardDataStub[0].langauge,
+    keymanKeyboardDataStub[0].bcp47Tag
+  );
+
+  t.is(await db.keyboardData.count(), 1);
+});
+
+test("retrieve Keyman keyboard data with addKeybaordData", async (t) => {
```
eddieantonio: Typos:

```suggestion
test("save Keyman keyboard data with addKeyboardData", async (t) => {
  const { db, storage } = t.context;

  t.is(await db.keyboardData.count(), 0);
  await storage.addKeyboardData(
    keymanKeyboardDataStub[0].langauge,
    keymanKeyboardDataStub[0].bcp47Tag
  );

  t.is(await db.keyboardData.count(), 1);
});

test("retrieve Keyman keyboard data with addKeyboardData", async (t) => {
```

eddieantonio (changes requested):
A few more refinements and typos to fix before it can be approved. The end is in sight!

IanSeng:
@eddieantonio  I not sure why GitHub does not allow me to reply to your comments I am going to do it here
----
- Q: After reading the documentation, I feel like ?q=l returns all keyboards whose languages have an "l" in the name. I think ?q= will return ALL keyboards. 
- A: You are absolutely correct. I double checked with the doc and compared the returning object from both `q=` and `q=l`, and `q=` is the correct one. 

----
- Q: Every time the worker starts up, it will fetch the entire language list. This is bad because it should really only be fetched maybe once a day, at most. But if you open a new tab, a new worker thread starts, which issues this request to the Keyman API 💀

Is there a way you can set a limit to how often the data is fetched?
- A: A new issue has been created for it. Link: [https://github.com/eddieantonio/predictive-text-studio/issues/135](https://github.com/eddieantonio/predictive-text-studio/issues/135 )

----
- Q: I'm still waiting for an explaination for this one. Is it intended that when you call fetchLanguageData() more than once, that the languageArray may potentially have many duplicate languages?

- A: this.langageArray = []; (after refactor - **languages: KeyboardData[]**;) was just to instantiate the array when the class is called. I could be doing it wrong? 

----
The rest of the comments have been addressed accordingly.

src/worker/keyman-api-service.ts
```diff
@@ -0,0 +1,42 @@
...
+  /**
+   * API configs
+   */
+  baseUrl = "https://api.keyman.com/search";
+  /**
+   * Query all languages data, doc: https://help.keyman.com/developer/cloud/search/1.0/
+   */
+  params = { q: "" };
+  url: URL;
+  languages: KeyboardData[];
```
eddieantonio: My issue with this is why is this an instance property? What purpose does it serve as an instance property? Why _must_ it be more than a local variable in the scope of `fetchLanguageData()`?

IanSeng:
@eddieantonio I misunderstood your comment earlier and have moved the variable to local in the scope of `fetchLanguageData()`

eddieantonio (approved):
LGTM!

PR #126, opened 2020-11-03
change: clean up and remove deprecated/duplicated code

eddieantonio:
Tidies up the codebase, removing code that is now unused, or used only in tests.

I also snuck in a change to remove checking if project data is undefined.


xqiu1 (approved):
LGTM!

Issue #127, opened 2020-11-04
Create a split button component


xqiu1:
Create a split button component for LanguageSource page
This issue is related to #124

xqiu1:
refactor <SplitButton> component so that it can be composed of two or more regular buttons

Issue #37, opened 2020-09-19
Add support to enter word and count manually


jienius:
* Users should be able to create a 'Direct Entry' source.

* When users select 'Direct Entry' as their source type, a table should appear with three columns (word, word count, actions) and one 'new row'.

* Each row will have one input field for the word, one optional input field for the word count, and one 'delete' icon.

  * Users can edit values for existing rows by selecting the input field.

  * When users select a row, the styling of the row should change to indicate it has been selected.

  * Users can type in a word into a new row and when they do another new row should appear below automatically.

  * Words have to be strings and cannot be empty.

  * Word counts have to be > 0 or empty.

  * Clicking the delete icon will remove the appropriate row from the table.

 * Users can enter a name for the Direct Entry Table via a text input field.

* The table will automatically save periodically and will have a manual save button.

* The table has to have at least one valid row to be saved or added.

xqiu1:
It's related to issue #124

PR #128, opened 2020-11-05
Add features for language sources page

### LINKED ISSUE #124: Add feature to add source from the Language Sources page

xqiu1:
Fixes #124


xqiu1:
Do I need to modify handleEdit and handleDelete(both methods in the LanguageSource page) in this pr?
PS: This is not a complete pr

eddieantonio:
> Do I need to modify handleEdit and handleDelete(both methods in the LanguageSource page) in this pr?
> PS: This is not a complete pr

Strictly speaking, you only need to be able to upload from the language sources page. So no, you don't _have_ to!

src/app/components/ManualEntry.svelte
```diff
@@ -0,0 +1,135 @@
...
+            </td>
+            <td>
+                <input type="number" placeholder="Optional" />
+            </td>
+            <td>
+                <button class="btn--inline" on:click={handleDelete}>Delete</button>
+            </td>
+        </tr>
+  </table>
+</div>
```
xqiu1, line 133: I find there is another issue #37 relates to this component. So in this pr, I will only do a general one. Later, I will finish the rest functionalities in a new pr for #37.

cypress/integration/addSource.spec.ts
```diff
@@ -0,0 +1,19 @@
+describe("Add source feature in LanguageSource page", function () {
+  it("should find a button to press to add source by uploading file", function () {
+    cy.visit("/languages");
+    cy.get("button").contains("Sources").click();
+
+    // Add source component should show after clicking the Add Source Button
+    cy.get("button").contains("Add Source").click().scrollIntoView();
+    cy.get("[data-cy=add-source-window]").should("exist");
+
+    cy.get("button").contains("UPLOAD").click();
```
IanSeng: Since you are already using `data-cy` attribute throughout your test, I would suggest adding it to the button as well.

src/app/components/AddSource.svelte
```diff
@@ -0,0 +1,183 @@
...
+  import SplitButton from "./SplitButton.svelte";
+  import ManualEntry from "./ManualEntry.svelte";
+  const UPLOAD_INPUT_ID = "upload-input";
+
+  const leftText = "UPLOAD";
+  const rightText = "DIRECT ENTRY";
+  export let addSource: boolean = false;
+  let manualEntry: boolean = false;
+  let onDraggedOver = false;
+  let fileList: any = [];
```
IanSeng: would avoid using `any` if there is a type for the `fileList` array

src/app/components/AddSource.svelte
```diff
@@ -0,0 +1,183 @@
...
+        const file = item.getAsFile();
+        if (file !== null) {
+          fileList.push(file);
+        }
+      }
+    }
+    return fileList;
+  }
+
+  const handleDrop = async (event: DragEvent) => {
```
IanSeng:
```suggestion
  const handleDrop = async (event: DragEvent) => {
```
```suggestion
  const handleDrop = (event: DragEvent) => {
```
`Async` can be removed here if there is no `await` in the code

src/app/components/AddSource.svelte
```diff
@@ -0,0 +1,183 @@
...
+    flex-direction: column;
+    align-items: center;
+    justify-content: center;
+    padding: 1%;
+  }
+</style>
+
+{#if addSource}
+  <div class="arrow-button-zone" data-cy="add-source-window">
+    <img src="assets/down-arrow.svg" alt="" role="presentation" class="block" />
```
IanSeng:
```suggestion
    <img src="assets/down-arrow.svg" alt="" role="presentation" class="block" />
```
```suggestion
    <img src="assets/down-arrow.svg" alt="Down Arrow" role="presentation" class="block" />
```
Adding alternative text for images for the reason of web accessibility. A good reference that I personally use https://webaim.org/techniques/alttext/

src/app/components/LanguageSources.svelte
```diff
@@ -15,7 +17,8 @@
    * @return {void}
    */
   const handleAddSource = (): void => {
-    console.log("Add source button clicked");
+    addSource = true;
+    console.log("Add button clicked");
```
IanSeng: Is this `console` needed or just for debugging?

src/app/components/ManualEntry.svelte
```diff
@@ -0,0 +1,135 @@
+<script lang="ts">
+  /**
+   * Handles the click when the delte button is pressed
+   * TODO
+   */
+  const handleDelete = (): void => {
+    console.log("Delete button clicked");
```
IanSeng: Is this `console` needed or just for debugging?

src/app/components/ManualEntry.svelte
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  .language__sources-manual-entry-tablename input {
+    height: 30px;
+    width: 400px;
+    border-radius: 2px;
+    border: 1px solid var(--lite-white);
+    padding: 5px;
+    text-align: center;
+    font: 24px var(--main-font);
```
IanSeng: Just a FYI. A global font size variables have been introduced in this pr [#98 ](https://github.com/eddieantonio/predictive-text-studio/pull/131/files/9759ce99024d8cfc87b32a60e1bcf353dfe3a781#diff-7e6d987ee4579aedab388447b754b3ba599ac9051c5c9c16a82cc59367ed64de).

cypress/integration/manualEntry.spec.ts
```diff
@@ -0,0 +1,25 @@
+describe("", function () {
+  it("should find a button to press to add source by user entry manually", function () {
```
IanSeng, line 2: I realized that `cypress/integration/addSource.spec.ts` and `cypress/integration/manualEntry.spec.ts` are both testing the components on the same page (`/languages`), and I am wondering why did you create two different test files?

IanSeng (changes requested):
I cannot really comment on the logic as I did not know much about the task and there is not much of background/description about it. Just a few minor changes and it should be good to go

src/app/components/AddSource.svelte
xqiu1: Sure, since fileList will be assigned to different data types, I change its type to File[]|FileList.

src/app/components/LanguageSources.svelte
xqiu1: console is just for debugging

cypress/integration/manualEntry.spec.ts
xqiu1: Since they are for different components, like AddSource has the test addSource.spec.ts, ManualEntry has the test manualEntry.spec.ts. If we don't want tests for components, I can put these two together. I just think that we were testing components instead of the entire page, as we did before.

cypress/integration/addSource.spec.ts
xqiu1: Some components created by us, such as Button, data-cy is an unknown attribute, so it might be easier to select using "button"?

eddieantonio: I agree; please add the `data-cy` attribute and use it to select in the Cypress test. See here: https://docs.cypress.io/guides/references/best-practices.html#Selecting-Elements

IanSeng: Yup, you can add testing attribute to the reusable button component instead of just selecting selecting the button.

xqiu1: Sure!

src/app/components/LanguageSources.svelte
IanSeng: since this is a production code, I think it is a good practice to not include debugging console log as user do not have to know this info

cypress/integration/manualEntry.spec.ts
IanSeng: Personally, I would say it is good to have just one since you only have a total of two  test cases for both components, but you can definitely make the call.

src/app/components/AddSource.svelte
IanSeng: If fileList could be assigned to different type, you could provide the known type as well as any. 
Example:
`let example: string | any`

IanSeng (approved):
most comments have been addressed accordingly and the rest LGTM

src/app/components/LanguageSources.svelte
xqiu1: Okay, I will also remove the console in other places.

src/app/components/AddSource.svelte
```diff
@@ -0,0 +1,189 @@
+<script lang="ts">
+  import worker from "../spawn-worker";
+  import Button from "./Button.svelte";
+  import SplitButton from "./SplitButton.svelte";
+  import ManualEntry from "./ManualEntry.svelte";
+  const UPLOAD_INPUT_ID = "upload-input";
+
+  const leftText = "UPLOAD";
+  const rightText = "DIRECT ENTRY";
```
eddieantonio: I have nuanced comments about this. Constants are good for software maintainability, but why are they good? Constants define one single place for some content to be defined. This is useful when:

 1.  the same value gets used in multiple different places (don't repeat yourself, easier to change the value throughout the entire codebase if needed, since you only have to change it in one place); or
 2. the constant, such as a integer, declares a _non-obvious_ value — then you can give it a meaningful name (e.g., `WEEKS_IN_A_YEAR` is more easy to understand than `52` in a codebase)!
 3. The use of the constant in one context is related to its use in another context. It helps signify to the programmer which values are related.

See: `UPLOAD_INPUT_ID` — it has a non obvious value that may need to be changed in the future, and is used in at least two places in the source code that have a related purpose.

These does not fit either of the above criteria.

The names `leftText` and `rightText`, in my opinion, _obscure_ the meaning of "upload" and "direct entry". And the value is a user-readable label that makes sense more sense in its true context.

As such, delete these here...

```suggestion
```

...then see a following comment.

src/app/components/AddSource.svelte
```diff
@@ -0,0 +1,189 @@
...
+{#if addSource}
+  <div class="arrow-button-zone" data-cy="add-source-options">
+    <img
+      src="assets/down-arrow.svg"
+      alt="Down Arrow"
+      role="presentation"
+      class="block" />
+    <SplitButton
+      {leftText}
+      {rightText}
```
eddieantonio:
```suggestion
      leftText="Upload"
      rightText="Direct entry"
```

src/app/components/AddSource.svelte
```diff
@@ -0,0 +1,189 @@
...
+  }
+</style>
+
+{#if addSource}
+  <div class="arrow-button-zone" data-cy="add-source-options">
+    <img
+      src="assets/down-arrow.svg"
+      alt="Down Arrow"
+      role="presentation"
+      class="block" />
```
eddieantonio: Regarding accessibility, every image that is content must have a non-empty `alt=` tag. However, some images are just visual decoration, and the app would make sense if the image was removed entirely. Decorative images should have `alt="" role="presentation"`.

**You cannot have a non-empty `alt=""` attribute plus `role="presentation"`**!

So, if the image is just a **decoration**, have this:

```suggestion
    <img
      src="assets/down-arrow.svg"
      alt=""
      role="presentation"
      class="block" />
```

However, if the image has **useful content** that _should_ be read aloud to the user, then it should look like this:

```suggestion
    <img
      src="assets/down-arrow.svg"
      alt="Down arrow"
      class="block" />
```

src/app/components/Button.svelte
```diff
@@ -1,12 +1,11 @@
 <script lang="ts">
-  import { beforeUpdate } from "svelte";
-
   // props
   export let color: string = "grey"; // optional
   export let isOutlined: boolean = false; // optional
   export let hasDropShadow: boolean = false; // optional
   export let size: string = "medium"; // optional
   export let subtext: string = ""; // optional
+  export let data_cy: string = ""; // optional
```
eddieantonio: Please use `camelCase` when naming new variables in JavaScript.

```suggestion
  export let dataCy: string = ""; // optional
```

src/app/components/Button.svelte
```diff
@@ -30,7 +29,8 @@
     class="button button--{color} button--{size}"
     class:button--shadow={hasDropShadow}
     class:button--outline={isOutlined}
-    on:click={onClick}><slot /></button>
+    on:click={onClick}
+    data-cy={data_cy}><slot /></button>
```
eddieantonio:
```suggestion
    data-cy={dataCy}><slot /></button>
```

src/app/components/LanguageSources.svelte
```diff
@@ -135,7 +133,13 @@
...
-  <Button color="blue" isOutlined size="large" onClick={handleAddSource}>
+  <Button
+    color="blue"
+    isOutlined
+    size="large"
+    onClick={handleAddSource}
+    data_cy="language-sources-add-source">
     Add Source
   </Button>
+  <AddSource bind:addSource />
```
eddieantonio: This can be changed in a later PR, but I realize now that the better markup for this is [the `<details>`/`<summary>` elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details).

Using `<details>`/`<summary>` means you would not even need to have the `addSource` variable to begin with, plus you get accessibility for free!

src/app/components/ManualEntry.svelte
```diff
@@ -0,0 +1,133 @@
+<script lang="ts">
+  /**
+   * Handles the click when the delte button is pressed
+   * TODO
+   */
+  const handleDelete = (): void => {};
```
eddieantonio: In general, I'd like the names of these callbacks to be more meaningful than `handleX`. I'd rather they specificy _exactly_ what they do, e.g., `deleteRow`.

```suggestion
   * Handles the click when the delete button is pressed
   * TODO
   */
  const handleDelete = (): void => {};
```

src/app/components/ManualEntry.svelte
```diff
@@ -0,0 +1,133 @@
...
+    padding: 5px;
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    justify-content: center;
+  }
+
+  .language__sources-manual-entry-tablename input {
+    height: 30px;
+    width: 400px;
```
eddieantonio: Please follow the guidelines set here: https://github.com/eddieantonio/predictive-text-studio/blob/production/docs/conventions.md#css-prefer-rem-to-px-for-positioning and here: https://github.com/eddieantonio/predictive-text-studio/blob/production/docs/conventions.md#css-prefer-em-to-px-for-font-size

Please do not copy-paste directly from Figma! But definitely use Figma as a guide 😅

src/app/components/ManualEntry.svelte
```diff
@@ -0,0 +1,133 @@
...
+    margin-bottom: 50px;
+    box-shadow: 0px 25px 40px 0px rgba(0, 0, 0, 0.1);
+  }
+
+  tr {
+    border-bottom: 1pt solid var(--lite-white);
+  }
+
+  td {
+    padding: 10px 0px 10px 20px;
```
eddieantonio: Please follow the guidelines set here:  https://github.com/eddieantonio/predictive-text-studio/blob/production/docs/conventions.md#css-omit-the-unit-from-zero

src/app/components/SplitButton.svelte
```diff
@@ -0,0 +1,51 @@
...
+  // props
+  export let colorLeft: string = "blue"; // optional
+  export let colorRight: string = "grey"; // optional
+  export let isOutlined: boolean = false; // optional
+  export let hasDropShadow: boolean = false; // optional
+  export let size: string = "medium"; // optional
+  export let leftText: string;
+  export let rightText: string;
+  export let data_cy_left: string;
+  export let data_cy_right: string;
```
eddieantonio: Please use `camelCase` for these variables

src/app/components/SplitButton.svelte
```diff
@@ -0,0 +1,51 @@
...
+    width: 10rem;
+  }
+  button:first-child {
+    border-top-left-radius: 0.625em;
+    border-bottom-left-radius: 0.625em;
+  }
+  button:last-child {
+    border-left: 0em;
+    border-top-right-radius: 0.625em;
+    border-bottom-right-radius: 0.625em;
```
eddieantonio: I like this border rounding effect!

```suggestion
    border-top-left-radius: 0.625em;
    border-bottom-left-radius: 0.625em;
  }
  button:last-child {
    border-top-right-radius: 0.625em;
    border-bottom-right-radius: 0.625em;
```

src/app/components/SplitButton.svelte
```diff
@@ -0,0 +1,51 @@
...
+  .split-button-layout {
+    display: inline-flex;
+    padding: 3%;
+  }
+  button {
+    margin: 0em;
+    border-radius: 0rem;
+    float: left;
+    width: 10rem;
+  }
```
eddieantonio: Why do you have both `inline-flex` and `float` in these declarations? I think the same layout can be achieved by using `display: flex` and removing the `float` declaration.

Also, please follow this document! https://github.com/eddieantonio/predictive-text-studio/blob/production/docs/conventions.md#css-omit-the-unit-from-zero

eddieantonio (changes requested):
I would like to see some further refinement in this PR.

Additionally, I think there can be quite a lot of refactoring done, but let's put that off for future PRs for now, because I think this PR has gotten large enough. Here are list of possible refactorings:

 * factor out upload logic so that both landing page and this page can share logic
 * refactor `<SplitButton>` component so that it can be _composed_ of two or more regular buttons

---

Also note, that if you include the `button` class in an element, its text automatically gets capitalized:

https://github.com/eddieantonio/predictive-text-studio/blob/8ba800854e98df3f79b25787c602448c4891de6d/public/global.css#L61

so it's preferable that button labels get regular English case in the source code (e.g., `Direct Entry` instead of `DIRECT ENTRY`), and then, since the all-caps is a _design_ choice, let the CSS specifiy that the text gets capitalized using `text-transform`.

src/app/components/LanguageSources.svelte
eddieantonio: See #137 for a follow-up on this disucssion.

cypress/integration/addSource.spec.ts
```diff
@@ -0,0 +1,17 @@
...
+    cy.visit("/languages");
+    cy.get("[data-cy=languages-sources-btn]").contains("Sources").click();
+
+    // Add source component should show after clicking the details element
+    cy.get("[data-cy=language-sources-add-sources]").click().scrollIntoView();
+
+    cy.get("[data-cy=add-sources-splitbtn-left]").contains("Upload").click();
+    cy.get("[data-cy=add-source-upload-zone]").contains("label", "Browse file");
+
+    // After clicking the details element, the add source zone will gone
```
eddieantonio: I'm confused by this comment... Could you please rewrite it or delete it?

cypress/integration/manualEntry.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+      .click()
+      .scrollIntoView();
+    cy.get("[data-cy=manual-entry-input-tablename]").type("Common Greetings");
+    cy.get("[data-cy=manual-entry-input-word]").type("Hello");
+    cy.get("[data-cy=manual-entry-input-count]").type("112");
+
+    cy.get("[data-cy=manual-entry-delete]").contains("Delete").click();
+    //TODO: Current row should be deleted
+
+    // After clicking the details element, the add source zone will gone
```
eddieantonio: Here too:

src/app/components/AddSource.svelte
```diff
@@ -0,0 +1,169 @@
...
+<div class="save-zone">
+  <Button
+    color="blue"
+    isOutlined
+    size="large"
+    onClick={saveAddedSources}
+    dataCy="add-sources-save-btn">
+    Save
+  </Button>
+</div>
```
eddieantonio, line 169: I think this is good for the current PR, but it feels like a little too much for a component. Perhaps a future PR can refactor to break this down into smaller components! I think there's overlap with #127

src/app/components/ManualEntry.svelte
eddieantonio: I will do a more thorough review in the next PR :)

eddieantonio (approved):
LGTM! There are a few comments in the cypress tests that confuse me, but other than than, looks good!

PR #129, opened 2020-11-05
[#31] feat: Add support for uploading file via Google Sheet

### LINKED ISSUE #31: Add support for uploading file via Google Sheet

pranavnarang:
* Added feature where users can input a url for a Google Sheet and its values are read, stored, and used to generate a kmp. 
* Here is a sample sheet that can be used to see how the process works: https://docs.google.com/spreadsheets/d/1lzHxoMWHpdGecby4d0y15AVf80csA7iNkaMMEL54Q7g/edit#gid=0
  * In case that url is not accessible, this one should work: https://docs.google.com/spreadsheets/d/1lzHxoMWHpdGecby4d0y15AVf80csA7iNkaMMEL54Q7g/edit?usp=sharing
* Resolves #31 

* NOTE: The necessary env variables (client id and api key) need to be added to the sever used by Vercel before this will work on it.


src/app/components/GoogleSheetsInput.svelte
```diff
@@ -0,0 +1,179 @@
...
+    error = null;
+    gapi.client
+      .init({
+        apiKey: API_KEY,
+        clientId: CLIENT_ID,
+        discoveryDocs: DISCOVERY_DOCS,
+        scope: SCOPES,
+      })
+      .then(
+        function () {
```
IanSeng:
```suggestion
        ()=> 
```
Since you are resolving a promise here, using a `then catch` would make the code more descriptive

src/app/components/GoogleSheetsInput.svelte
```diff
@@ -0,0 +1,179 @@
...
+          // Listen for sign-in state changes.
+          gapi.auth2.getAuthInstance().isSignedIn.listen(updateSigninStatus);
+
+          // Handle the initial sign-in state.
+          updateSigninStatus(gapi.auth2.getAuthInstance().isSignedIn.get());
+        },
+        function (error) {
```
IanSeng:
```suggestion
        catch(err => { ... })
```

src/app/components/GoogleSheetsInput.svelte
```diff
@@ -0,0 +1,179 @@
...
+  import worker from "../spawn-worker";
+
+  import InputField from "./InputField.svelte";
+  import Button from "./Button.svelte";
+
+  var spreadsheetId = "";
+  var error = null;
+
+  // Client ID and API key from the Developer Console
+  var CLIENT_ID = process.env.G_CLIENT_ID;
```
IanSeng:
```suggestion
  const CLIENT_ID = process.env.G_CLIENT_ID;
```
`const` would be more appropriate here since we do not have this variable to be modified later. Same for the **API_KEY**, **DISCOVERY_DOCS** and **SCOPES**

src/app/components/GoogleSheetsInput.svelte
```diff
@@ -0,0 +1,179 @@
...
+                let wordCount = row[1];
+                if (!wordCount) {
+                  wordCount = 0;
+                }
+                wordListObject.push([word, wordCount]);
+              }
+              error = null;
+              console.log(wordListObject); // TODO: remove me
+
+              const kmpFile = await worker.readGoogleSheet(
```
IanSeng: this local constant `kmpFile` seems to be used here, if this line was intended to call readGoogleSheet function, const kmpFile can be removed 
```suggestion
              await worker.readGoogleSheet(
```

rollup.config.js
```diff
@@ -26,6 +31,11 @@ const appConfiguration = {
     file: "public/app.js",
   },
   plugins: [
+    replace({
+      "process.env.G_CLIENT_ID": JSON.stringify(process.env.CLIENT_ID),
+      "process.env.G_API_KEY": JSON.stringify(process.env.API_KEY),
+    }),
```
eddieantonio, line 37: This is really cool!

src/app/components/GoogleSheetsInput.svelte
```diff
@@ -0,0 +1,179 @@
...
+  var DISCOVERY_DOCS = [
+    "https://sheets.googleapis.com/$discovery/rest?version=v4",
+  ];
+
+  // Authorization scopes required by the API; multiple scopes can be
+  // included, separated by spaces.
+  var SCOPES = "https://www.googleapis.com/auth/spreadsheets.readonly";
+
+  // let authorizeButton = document.getElementById("authorize_button");
+  // let signoutButton = document.getElementById("signout_button");
```
eddieantonio:
```suggestion
```

src/app/components/GoogleSheetsInput.svelte
```diff
@@ -0,0 +1,179 @@
...
+      .then(
+        function () {
+          // Listen for sign-in state changes.
+          gapi.auth2.getAuthInstance().isSignedIn.listen(updateSigninStatus);
+
+          // Handle the initial sign-in state.
+          updateSigninStatus(gapi.auth2.getAuthInstance().isSignedIn.get());
+        },
+        function (error) {
+          error = JSON.stringify(error, null, 2);
```
eddieantonio: `error` _shadows_ the assignment in the outer-scope (`var error = null` near the top of the file`). So there's actually two different variables, both called `error` and this line is assigning to the inner one—not the intended `error` variable.

Regardless, there should probably be a better thing to do with the error from the API than stringifying and displaying that raw error message to the user. Perhaps a message like "Could not connect to Google Sheets"?

src/app/components/GoogleSheetsInput.svelte
```diff
@@ -0,0 +1,179 @@
...
+   */
+  function getSpreadsheetId() {
+    let url = document.getElementById("googleSheetsURL").value;
+    if (url) {
+      return url.split("/")[5];
+    } else {
+      error =
+        "Error: The url " +
+        url +
+        " is not a valid Google Sheets URl, please paste a valid one.";
```
eddieantonio:
```suggestion
      error = `Error: The url ${url} is not a valid Google Sheets URl. Please paste a valid one.`;
```

src/app/components/GoogleSheetsInput.svelte
```diff
@@ -0,0 +1,179 @@
...
+<pre id="content" style="white-space: pre-wrap;" />
+<div class="google-sheets">
+  {#if error}
+    <p class:error>{error}</p>
+  {/if}
+  <h2>Or Get Values from a Google Sheet</h2>
+  <InputField
+    label="Google Sheets URL"
+    id="googleSheetsURL"
+    value={spreadsheetId}
```
eddieantonio:
```suggestion
    bind:value={googleSheetsURL}
```

src/app/components/GoogleSheetsInput.svelte
```diff
@@ -0,0 +1,179 @@
...
+                let row = range.values[i];
+                let word = row[0];
+                let wordCount = row[1];
+                if (!wordCount) {
+                  wordCount = 0;
+                }
+                wordListObject.push([word, wordCount]);
+              }
+              error = null;
+              console.log(wordListObject); // TODO: remove me
```
eddieantonio:
```suggestion
```

src/app/components/GoogleSheetsInput.svelte
```diff
@@ -0,0 +1,179 @@
+<script>
+  import worker from "../spawn-worker";
+
+  import InputField from "./InputField.svelte";
+  import Button from "./Button.svelte";
+
+  var spreadsheetId = "";
+  var error = null;
```
eddieantonio, line 7: You'll need to add a line to do data binding (see below):

```suggestion
  var error = null;
  let googleSheetsURL = "";
```

src/app/components/GoogleSheetsInput.svelte
```diff
@@ -0,0 +1,179 @@
...
+  function handleSignoutClick(event) {
+    gapi.auth2.getAuthInstance().signOut();
+  }
+
+  /**
+   * Gets the unique spreadsheet ID from a Google Sheet URL
+   *
+   */
+  function getSpreadsheetId() {
+    let url = document.getElementById("googleSheetsURL").value;
```
eddieantonio:
```suggestion
    let url = googleSheetsURL;
```

src/app/components/GoogleSheetsInput.svelte
```diff
@@ -0,0 +1,179 @@
...
+  }
+
+  /**
+   * Read word and count form storage
+   * Sample: https://docs.google.com/spreadsheets/d/1lzHxoMWHpdGecby4d0y15AVf80csA7iNkaMMEL54Q7g/edit#gid=0
+   */
+  async function getValuesFromSpreadSheet() {
+    const wordListObject = [];
+    spreadsheetId = getSpreadsheetId() || "";
+    if (!error) {
```
eddieantonio: Suggest refactoring this to use a guard clause, so that the "meat" of this function becomes unindented by one level. It's ever so slightly cleaner!

Here are a few tutorials:
 - https://www.catalyst.net.nz/blog/early-returnsguard-clauses-javascript-and-react
 - https://learningactors.com/javascript-guard-clauses-how-you-can-refactor-conditional-logic/

src/app/components/GoogleSheetsInput.svelte
```diff
@@ -0,0 +1,179 @@
...
+  }
+
+  .error {
+    background-color: #f8d7db;
+    color: #400000;
+    padding: 5px 10px;
+  }
+</style>
+
+<pre id="content" style="white-space: pre-wrap;" />
```
eddieantonio: What is the purpose of this `<pre>`?

src/worker/predictive-text-studio-worker-impl.ts
```diff
@@ -21,6 +22,14 @@ export class PredictiveTextStudioWorkerImpl
     return await linkStorageToKmp(this.storage);
   }
 
+  async readGoogleSheet(
+    name: string,
+    wordListObject: WordList
+  ): Promise<ArrayBuffer> {
+    this.storage.saveFile(name, wordListObject);
+    return await linkStorageToKmp(this.storage);
```
eddieantonio, line 76: This behaviour will change when merged with `production`, but it's okay for this PR 😅

src/app/components/GoogleSheetsInput.svelte
eddieantonio: A disturbing thing here is that this should fail type-checking. The `PredIctiveTextWorker` interface does not have the method `readGoogleSheet()` (the concrete implementation has it though). ¯\\\_(ツ)\_/¯

eddieantonio (changes requested):
I think some things could be tightened up here. Also, do you mind DMing me on Slack so that we can setup the Google Sheets environment variables on the Vercel backend?

src/common/predictive-text-studio-worker.ts
```diff
@@ -1,11 +1,18 @@
...
+
 export interface PredictiveTextStudioWorker {
   /**
    * Save a file into IndexedDB
    */
   saveFile(name: string, file: File): Promise<ArrayBuffer>;
+  /**
+   * Save a google sheet into IndexedDB
+   */
+  readGoogleSheet(name: string, wordListObject: WordList): Promise<ArrayBuffer>;
```
```suggestion
  readGoogleSheet(name: string, wordList: WordList): Promise<ArrayBuffer>;
```

src/app/components/GoogleSheetsInput.svelte
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  .error {
+    background-color: #f8d7db;
+    color: #400000;
+    padding: 5px 10px;
+  }
+</style>
+
+<!-- <pre id="content" style="white-space: pre-wrap;" /> -->
```
eddieantonio:
```suggestion
```

src/app/components/GoogleSheetsInput.svelte
```diff
@@ -0,0 +1,172 @@
...
+    label="Google Sheets URL"
+    id="googleSheetsURL"
+    bind:value={googleSheetsURL}
+    fullWidth={true} />
+  <Button color="blue" onClick={handleClientLoad}>Read Values</Button>
+</div>
+<svelte:head>
+  <script async defer src="https://apis.google.com/js/api.js">
+  </script>
+</svelte:head>
```
eddieantonio: I would prefer `<svelte:head>` go above the main HTML content (above the `<div>`)

src/app/components/GoogleSheetsInput.svelte
```diff
@@ -0,0 +1,172 @@
...
+          // Listen for sign-in state changes.
+          gapi.auth2.getAuthInstance().isSignedIn.listen(updateSigninStatus);
+
+          // Handle the initial sign-in state.
+          updateSigninStatus(gapi.auth2.getAuthInstance().isSignedIn.get());
+        },
+        function () {
+          error = "Error: Could not connect to Google Sheets";
+        }
+      );
```
eddieantonio: This is more conventional style for Promise-based API usage:

```suggestion
      .then(() => {
        // Listen for sign-in state changes.
        gapi.auth2.getAuthInstance().isSignedIn.listen(updateSigninStatus);
        // Handle the initial sign-in state.
        updateSigninStatus(gapi.auth2.getAuthInstance().isSignedIn.get());
      })
      .catch(() => {
        error = "Error: Could not connect to Google Sheets";
      });
```

src/app/components/GoogleSheetsInput.svelte
```diff
@@ -0,0 +1,172 @@
...
+    } else {
+      handleAuthClick();
+      getValuesFromSpreadSheet();
+    }
+  }
+
+  /**
+   *  Sign in the user upon button click.
+   */
+  function handleAuthClick(event) {
```
eddieantonio: When possible, prefer function names that describe the _purpose_ of what they do, rather than `handleEVENT` names.

```suggestion
  function signInToGoogleAPI(event) {
```

src/app/components/GoogleSheetsInput.svelte
```diff
@@ -0,0 +1,172 @@
...
+    gapi.auth2.getAuthInstance().signIn();
+  }
+
+  /**
+   *  Sign out the user upon button click.
+   */
+  function handleSignoutClick(event) {
+    gapi.auth2.getAuthInstance().signOut();
+  }
```
eddieantonio: Delete all unused functions

```suggestion
```

src/app/components/GoogleSheetsInput.svelte
```diff
@@ -0,0 +1,172 @@
...
+    display: flex;
+    flex-direction: column;
+    justify-content: center;
+    align-items: center;
+  }
+
+  .error {
+    background-color: #f8d7db;
+    color: #400000;
+    padding: 5px 10px;
```
eddieantonio: Please use `em` or `rem` units instead (probably `rem`s). See: https://github.com/eddieantonio/predictive-text-studio/blob/production/docs/conventions.md#css-prefer-rem-to-px-for-positioning

src/app/components/GoogleSheetsInput.svelte
```diff
@@ -0,0 +1,172 @@
...
+  }
+</style>
+
+<!-- <pre id="content" style="white-space: pre-wrap;" /> -->
+<div class="google-sheets">
+  {#if error}
+    <p class:error>{error}</p>
+  {/if}
+  <h2>Or Get Values from a Google Sheet</h2>
```
eddieantonio: Are you using this `<h2>` for to affect the font-size, or are you using it as a secondary heading? If the latter, what are the other level two headings?

src/app/pages/LandingPage.svelte
```diff
@@ -208,6 +209,7 @@
 
       <button class="button button--primary button--shadow quick-start__submit" type="submit"> Upload </button>
     </form>
+    <GoogleSheetsInput />
```
eddieantonio, line 288: This is part of the `<form>` above, and should be placed _above_ the `<button type="submit">`

src/app/components/GoogleSheetsInput.svelte
```diff
@@ -0,0 +1,172 @@
...
+   * Read word and count form storage
+   * Sample: https://docs.google.com/spreadsheets/d/1lzHxoMWHpdGecby4d0y15AVf80csA7iNkaMMEL54Q7g/edit#gid=0
+   */
+  async function getValuesFromSpreadSheet() {
+    const wordListObject = [];
+    let spreadsheetId = getSpreadsheetId() || "";
+
+    if (error) {
+      return error;
+    }
```
eddieantonio: I'm confused about this. So if the global error is set, this function should return the error? So the return type of this functinon becomes `Promise<string> | Promise<void>`? In general, you want to keep the return type as simple as possible and "monomorphic" (one type rather than potentially many types).

src/app/components/GoogleSheetsInput.svelte
```diff
@@ -0,0 +1,172 @@
...
+              let word = row[0];
+              let wordCount = row[1];
+              if (!wordCount) {
+                wordCount = 0;
+              }
+              wordListObject.push([word, wordCount]);
+            }
+            error = null;
+
+            await worker.readGoogleSheet(spreadsheetId, wordListObject);
```
eddieantonio: I don't think this function needs to be awaited, unless you are explicitly handling exceptions from it.

src/app/components/GoogleSheetsInput.svelte
```diff
@@ -0,0 +1,172 @@
...
+          } else {
+            error = "Error: No data found in the Google Sheet.";
+          }
+        },
+        function (response) {
+          error = "Error: " + response.result.error.message;
+        }
+      );
+  }
```
eddieantonio: If you declare a function `async`, do not mix both `async/await` syntax with Promise callbacks. Instead, use `async`/`await` throughout:

```suggestion

  async function getValuesFromSpreadSheet() {
    const wordListObject = [];
    const spreadsheetId = getSpreadsheetId() || "";
    if (error) {
      return error;
    }

    let response;
    try {
      response = await gapi.client.sheets.spreadsheets.values.get({
        spreadsheetId: spreadsheetId,
        range: "A2:B",
      });
    } catch {
      error = "Error: " + response.result.error.message;
      return;
    }

    const range = response.result;
    if (range.values.length <= 0) {
      error = "Error: No data found in the Google Sheet.";
      return;
    }

    for (let i = 0; i < range.values.length; i++) {
      const row = range.values[i];
      const word = row[0];
      let wordCount = row[1];
      if (!wordCount) {
        wordCount = 0;
      }
      wordListObject.push([word, wordCount]);
    }
    error = null;

    await worker.readGoogleSheet(spreadsheetId, wordListObject);
  }
```

eddieantonio (changes requested):
I think this can be refined just a bit. Feel free to DM me if you have any further questions!

eddieantonio (approved):
LGTM!

I made some sneaky edits, but thanks for getting this done!

PR #134, opened 2020-11-09
change(style): always use T[] notation for array types

eddieantonio:
I keep complaining that `Array<SomeType>` ("generic style") should be written as `SomeType[]` ("array style"). This automatically enforces it, _and_ automatically fixes it too!


IanSeng (approved):
LGTM

PR #136, opened 2020-11-10
[#132] feat: Implement auto complete to the "Input field" component

### LINKED ISSUE #132: Implement auto complete to the "Input field" component
### Description 
- currently the input field does not support the functionally to display suggestions based on input text
- leverage on the pre-existing component `InputField.svelte` to improve the UI component

### Subtasks
- [x] input component with suggestion 
- [x] clickable and update the text in the input field 
- [x] component testing in cypress to ensure it work as expected

IanSeng:
# Summary 
This PR is to create a autocomplete component for the use of suggesting language and display the correct BCP47. 
- Since this component consists of multiple functionality, I have decided to make it a standalone component instead of leveraging existing input component to decrease the complexity. 
- Users are now able select suggestion list by mouse click and keyboard navigation
- Since all of the spacing and border pixel are the factor of 5, the style is extracted and a global variable is created for spacing and border in this pr. 

~~Note: Sorry for the inconvenience and confusion, I have by accident branched out from `feat/keyboard-data` instead of `Production`, thus this PR will goes in after [https://github.com/eddieantonio/predictive-text-studio/pull/123](url). Only the following files are touched in this PR:~~
~~- cypress/integration/languages.spec.ts~~
~~- src/app/components/AutoComplete.svelte~~
~~- src/app/pages/Languages.svelte~~
~~- public/global.css~~

**Note**: above is not applicable anymore since [#123](https://github.com/eddieantonio/predictive-text-studio/pull/123) has been merged into the `production` branch and this branch has been rebased with the latest changes.  

### Sub-tasks
- [x] input component with suggestion
- [x] clickable and update the text in the input field
- [x] component testing in cypress to ensure it work as expected
- [x] extract global spacing and border styles 
- [x] ensure a consistent design 

### Issue
Resolves #132


public/global.css
```diff
@@ -31,6 +32,15 @@
...
   --xxl: 2.5em;
+
+  /* Spacing and border variables */
+  --sb-xxs: 5px;
+  --sb-xs: 10px;
+  --sb-s: 15px;
+  --sb-m: 20px;
+  --sb-l: 25px;
+  --sb-xl: 30px;
+  --sb-xxl: 40px;
```
eddieantonio: Please use `rem` so that it scales with the user's default font size (accessibility). https://github.com/eddieantonio/predictive-text-studio/blob/production/docs/conventions.md#css-prefer-rem-to-px-for-positioning

eddieantonio (commented):
I'm gonna do a more thorough review of this shortly, but I started reviewing this yesterday and... forgot to submit this comment 😅

src/app/pages/Languages.svelte
```diff
@@ -41,6 +41,21 @@
       },
     ],
   };
+  // TODO: Temporary data placeholder for autocorrect component and will be removed once #33 is done
+  let testingData: any = [
```
eddieantonio, line 45: Avoid using `any`:

```suggestion
  let testingData: any = [
```

cypress/integration/languages.spec.ts
```diff
@@ -0,0 +1,64 @@
+describe("AutoComplete Component", () => {
+  beforeEach(() => {
+    cy.visit("/languages?");
```
eddieantonio: Why the empty query string?
```suggestion
    cy.visit("/languages");
```

cypress/integration/languages.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+  });
+
+  it("should display the suggestion list", () => {
+    getDataCY("autocomplete-input")
+      .type("e")
+      .then(() => {
+        getDataCY("autocomplete-suggestions").its("length").should("eq", 3);
+      });
```
eddieantonio: These tests depend HEAVILY on the sample automcompletion database. #150 will help use more predicitable and realistic test data in this test!

cypress/integration/languages.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  it("should display the correct BCP47 subtext", () => {
+    getDataCY("autocomplete-input")
+      .type("e")
+      .then(() => {
+        getDataCY("autocomplete-suggestions").first().click();
+        getDataCY("autocomplete-subtext").should(
+          "have.text",
+          "BCP47Tag:\n    en"
+        );
```
eddieantonio: The assertion as written is brittle, as if so much as a space or newline is added or deleted, it breaks this test. How about?

```suggestion
        getDataCY("autocomplete-subtext").should("contain", "en");
```

If I understand correctly, `contain` uses the `.contain(text)` assertion from [chai-jquery](https://github.com/chaijs/chai-jquery#containtext).

Alternatively, you could use a regular expression, but it's a bit more annoying to implement:

```suggestion
        getDataCY("autocomplete-subtext")
          .invoke("text")
          .should("match", /BCP-?47\s?Tag:\s+en\b/);
```

cypress/integration/languages.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+      });
+  });
+});
+
+function getDataCY(name: string) {
+  return cy.get(`[data-cy=${name}]`);
+}
```
eddieantonio: I really like the idea of this shortcut, but I think it'd make more sense to create a custom command that does this. Something like `cy.data(name)`... 🤔

src/app/components/AutoComplete.svelte
```diff
@@ -0,0 +1,188 @@
...
+  let selected: string = "";
+  // Toggle to show search list
+  let show = false;
+  // Index of Previous focus element
+  let prevIndex = -1;
+  // Index of focus element
+  let index = -1;
+
+  // Input to search
+  function onChange(event: any) {
```
eddieantonio: Try to avoid `any`; I think the type here is `Event`:

```suggestion
  function onChange(event: Event) {
```

src/app/components/AutoComplete.svelte
```diff
@@ -0,0 +1,188 @@
...
+    Up = "ArrowUp",
+    Down = "ArrowDown",
+    Enter = "Enter",
+  }
+  export let label = "";
+  export let subtext = "";
+  // Testing Data
+  export let results: dataObj[] = [];
+  // To store Filted array
+  let filetred: dataObj[] = [];
```
eddieantonio: Typo:

```suggestion
  // To store filtered array
  let filtered: dataObj[] = [];
```

It will also need to be changed throught the rest of the file!

src/app/components/AutoComplete.svelte
```diff
@@ -0,0 +1,188 @@
...
+  let show = false;
+  // Index of Previous focus element
+  let prevIndex = -1;
+  // Index of focus element
+  let index = -1;
+
+  // Input to search
+  function onChange(event: any) {
+    show = true;
+    filetred = results.filter((item: any) => {
```
eddieantonio:
```suggestion
    filetred = results.filter((item: dataObj) => {
```

src/app/components/AutoComplete.svelte
```diff
@@ -0,0 +1,188 @@
...
+  // On select item in list
+  function selectedList(data: dataObj) {
+    show = false;
+    selected = data.language;
+    subtext = data.bcp47Tag;
+  }
+
+  const mod = (a: number, n: number) => {
+    return a - n * Math.floor(a / n);
+  };
```
eddieantonio, line 67: Why is this used instead of `a % n`?

src/app/components/AutoComplete.svelte
```diff
@@ -0,0 +1,188 @@
...
+        prevIndex = index;
+        index -= 1;
+      }
+    } else if (key == keyboardKey.Enter) {
+      selected = filetred[index].language;
+      subtext = filetred[index].bcp47Tag;
+      show = false;
+    }
+    const allItems = [
+      ...document.getElementsByClassName("autocomplete__suggestion-item"),
```
eddieantonio: Oooo, this is very clever! I would have used `Array.from()` instead of using the splat, but I can't decide if this is more readable... leave it this way! I'll figure it out later!

src/app/components/AutoComplete.svelte
```diff
@@ -0,0 +1,188 @@
...
+        index = -1;
+      } else {
+        prevIndex = index;
+        index -= 1;
+      }
+    } else if (key == keyboardKey.Enter) {
+      selected = filetred[index].language;
+      subtext = filetred[index].bcp47Tag;
+      show = false;
+    }
```
eddieantonio: Remove the need for the first conditional by handling it in the `else` clause:

```suggestion
    if (key === keyboardKey.Down) {
      prevIndex = index;
      index += 1;
    } else if (key === keyboardKey.Up) {
      if (index == -1) {
        index = -1;
      } else {
        prevIndex = index;
        index -= 1;
      }
    } else if (key === keyboardKey.Enter) {
      selected = filetred[index].language;
      subtext = filetred[index].bcp47Tag;
      show = false;
    } else {
      /* Not a key we care about */
      return;
    }
```

You could even use a `switch`/`case`/`default` if you're so bold! 👀

src/app/components/AutoComplete.svelte
```diff
@@ -0,0 +1,188 @@
...
+      if (index == -1) {
+        index = -1;
+      } else {
+        prevIndex = index;
+        index -= 1;
+      }
+    } else if (key == keyboardKey.Enter) {
+      selected = filetred[index].language;
+      subtext = filetred[index].bcp47Tag;
+      show = false;
```
eddieantonio: Of course, fix the typo in the variable, but you can use the existing function!

```suggestion
      selectedList(filetred[index])
```

src/app/components/AutoComplete.svelte
```diff
@@ -0,0 +1,188 @@
...
+    cursor: pointer;
+    background-color: var(--gray-highlight);
+  }
+  .autocomplete__subtext {
+    font-family: var(--mono-font), monospace;
+    color: var(--gray);
+    font-size: var(--xxs);
+  }
+  /*In Sevelte, dynamic class will only work if it's a global style*/
+  :global(.autocomplete-active) {
```
eddieantonio: I'm not super sure what this comment is about, but here's an idea!

```suggestion
  .autocomplete--active {
```

src/app/components/AutoComplete.svelte
```diff
@@ -0,0 +1,188 @@
...
+    on:input={(event) => onChange(event)}
+    data-cy="autocomplete-input" />
+  {#if show}
+    <ul
+      class="autocomplete__suggestion-list"
+      use:clickOutside={closeSuggestion}>
+      {#each filetred as result}
+        <li
+          class="autocomplete__suggestion-item"
+          data-cy="autocomplete-suggestions">
```
eddieantonio: Use the index to match with the currently selected class (see: https://svelte.dev/tutorial/each-blocks)

```suggestion
      {#each filetred as result, i}
        <li
          class="autocomplete__suggestion-item"
          class:autocomplete--active={index == i}
          data-cy="autocomplete-suggestions"
```

src/app/components/AutoComplete.svelte
```diff
@@ -0,0 +1,188 @@
...
+
+    index = mod(index, allItems.length);
+    document
+      .getElementsByClassName("autocomplete__suggestion-item")
+      [index].classList.add("autocomplete-active");
+    if (prevIndex !== -1) {
+      document
+        .getElementsByClassName("autocomplete__suggestion-item")
+        [prevIndex].classList.remove("autocomplete-active");
+    }
```
eddieantonio: Using the `class:autocomplete--active={...}` trick in Svelte removes the need for this code.
```suggestion
```

I believe this also removes the need for the `prevIndex` variable.

eddieantonio (changes requested):
Good start, but I think this could use a bit more polishing!

cypress/integration/languages.spec.ts
eddieantonio: #154 lol

src/app/components/AutoComplete.svelte
IanSeng: omg! I wish I could Grammarly my code 😭😂

src/app/components/AutoComplete.svelte
IanSeng: javascript seems like it not smart enough to handle modulo with negative numerator, this function is to handle modules with both positive and negative numerator

src/app/components/AutoComplete.svelte
IanSeng: omg I love this refactor and it is way more readable than my previous code!!!

src/app/components/AutoComplete.svelte
IanSeng: this has essentially solved my dumb and ineffective trick of using global!!! thank you so much

src/app/pages/Languages.svelte
IanSeng: The testing data has been removed from code in this pr [#152](https://github.com/eddieantonio/predictive-text-studio/pull/152), since I did it in a backward manner by building the component first which why I implement a placeholder testing data. I am sorry for the confusion😔, I will also rebase #152 once this pr goes in so to ease out the pr review process and confusion.

cypress/integration/languages.spec.ts
IanSeng: I totally agree on your comment and have decided to completely remove all of these testings since they are very ineffective.

cypress/integration/languages.spec.ts
IanSeng: I love that implementation #154 !!!

src/app/components/AutoComplete.svelte
IanSeng: Same as above, the `any` has been fixed from code in this pr #152. sorry again for the confusion.

IanSeng:
> Good start, but I think this could use a bit more polishing!

- Thank you so much for the comments and I have addressed them accordingly

src/app/components/AutoComplete.svelte
eddieantonio: Ah, when index becomes `-1` `-1 % allItems.length === -1` where as `mod(-1, allItems.length) === allItems.length -1`. The latter is what we want! 👍🏼

src/app/components/AutoComplete.svelte
```diff
@@ -0,0 +1,176 @@
...
+      selectedList(filtered[index]);
+    } else {
+      /* Not a key we care about */
+      return;
+    }
+    const allItems = [
+      ...document.getElementsByClassName("autocomplete__suggestion-item"),
+    ];
+
+    index = mod(index, allItems.length);
```
eddieantonio: Divide by zero when `allItems` is empty. That _shouldn't_ happen, but just in case:

```suggestion
    index = allItems.length > 0 ? mod(index, allItems.length) : 0;
```

src/app/components/AutoComplete.svelte
```diff
@@ -0,0 +1,176 @@
...
+      }
+    } else if (key === keyboardKey.Enter) {
+      selectedList(filtered[index]);
+    } else {
+      /* Not a key we care about */
+      return;
+    }
+    const allItems = [
+      ...document.getElementsByClassName("autocomplete__suggestion-item"),
+    ];
```
eddieantonio: I just noticed that this clever trick isn't actually needed. `allItems` is created to get the `.length` to bound the modulus. However, [`document.getElementsByClassName()`](https://developer.mozilla.org/en-US/docs/Web/API/Document/getElementsByClassName) returns an `HTMLCollection`, which [has a `.length` property](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCollection#Properties)! The `allItems` array is not used otherwise, so just get its length!

 ```suggestion
    const numberOfSuggestions = document.getElementsByClassName("autocomplete__suggestion-item").length;
```

eddieantonio (approved):
LGTM!

There are a few minor improvements, but they're not worth blocking. Good work!

src/app/components/AutoComplete.svelte
IanSeng: I like this suggestion and it also omit the use of spread operator

IanSeng:
> LGTM!
> 
> There are a few minor improvements, but they're not worth blocking. Good work!

Thank you so much for the comments and I have also addressed them accordingly 🎉🎉🎉

PR #139, opened 2020-11-12
[#49] Update drag and drop cypress test

### LINKED ISSUE #49: Add Cypress test to test upload via drag-and-drop
This answer might be useful?

https://stackoverflow.com/questions/55361499/how-to-implement-drag-and-drop-in-cypress-test/55436989

Or it might not be useful lol

---

**EDIT**: this might be useful too: https://github.com/cypress-io/cypress-example-recipes/issues/173#issuecomment-635898328

jienius:
Fixes #49

### DIFF
--- a/src/app/components/Upload.svelte
+++ b/src/app/components/Upload.svelte
@@ -4,7 +4,7 @@
   import * as Comlink from "comlink";
   const UPLOAD_INPUT_ID = "upload-input";
 
-  let onDraggedOver = false;
+  let dragEnterCounter = 0;
   let downloadURL = "";
 
   worker.onPackageCompileSuccess(
@@ -33,7 +33,7 @@
   }
 
   const handleDrop = async (event: DragEvent) => {
-    onDraggedOver = false;
+    dragEnterCounter = 0;
     let fileList: File[] = [];
 
     if (event.dataTransfer == null) {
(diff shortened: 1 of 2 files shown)


src/app/components/Upload.svelte
```diff
@@ -110,10 +110,11 @@
 
 <div
   class="upload-zone"
-  class:drag-over={onDraggedOver}
+  class:drag-over={dragEnterCounter > 0}
   on:drop|preventDefault={handleDrop}
-  on:dragover|preventDefault={handleDragOver}
```
jienius, line 115: dragOver on chrome doesn't work very smooth

src/app/components/Upload.svelte
```diff
@@ -110,10 +110,11 @@
 
 <div
   class="upload-zone"
-  class:drag-over={onDraggedOver}
+  class:drag-over={dragEnterCounter > 0}
```
eddieantonio, line 121: This is fascinating! Could you please add a comment about this behaviour somewhere in the code with a link to the fact that Chrome is bad at dragover?

eddieantonio (approved):
Just one change (add a comment about why `dragEngerCounter` is necessary), and then LGTM!

PR #140, opened 2020-11-13
[#135] feat:  Enhance the efficiency by limiting the fetch amount

### LINKED ISSUE #135: Enhance the efficiency by limiting the fetch amount
# Description
Every time the worker starts up, it will fetch the entire language list. This is bad because it should really only be fetched maybe once a day, at most. But if you open a new tab, a new worker thread starts, which issues this request to the Keyman API 💀 - [@eddieantonio]

Is there a way you can set a limit to how often the data is fetched?

## Subtasks
- [x] Limit the number of time of fetching data 
- [x] If data is stored in the browser check last fetched date to determine if a new fetch is needed
- [x] Clear out the data before put to avoid redundancy of data
- [x] Unit testing 
- etc..

IanSeng:
# Summary
**Background**: This PR is to enhance/decrease the numbers of time fetch language list is called. More background can be found [here](https://github.com/eddieantonio/predictive-text-studio/issues/135)

**Enhancements Logic** : 
- Fetch language list from the Keyman API if there is np `keyboardData` data in indexeddb
- Fetch language list API will be called if the data saved in` indexeddb ` is more than **7 days** and the pre existing data will be clear out to avoid data duplication

## Subtasks
- [x] Limit the number of time of fetching data 
- [x] If data is stored in the browser check last fetched date to determine if a new fetch is needed (currently set to 7 days) 
- [x] Clear out the data before put to avoid redundancy of data
- [x] Unit testing 

**Note**: I have to admit that I did not do a great job in testing the `test-worker.ts` for this specific feature 😞 and I have also time boxed myself almost sending 6 hours in researching as well as trying to test logic with `sinon spy/ stub` function. Although, I wasn't able to cover some of the test cases in unit test but I did manually tested the logic with the browser dev tool🤓. Feel free to ping me if you have any idea to further improve the testing. 

## Resolves #
#135

### DIFF
--- a/src/worker/keyman-api-service.ts
+++ b/src/worker/keyman-api-service.ts
@@ -14,7 +14,7 @@ export class KeymanAPI {
   /**
    * Query all languages data, doc: https://help.keyman.com/developer/cloud/search/1.0/
    */
-  params = { q: "" };
+  params = { q: "l" };
   url: URL;
   constructor() {
     this.url = new URL(this.baseUrl);
--- a/src/worker/keyman-api-service.ts
+++ b/src/worker/keyman-api-service.ts
@@ -14,7 +14,7 @@ export class KeymanAPI {
   /**
    * Query all languages data, doc: https://help.keyman.com/developer/cloud/search/1.0/
    */
-  params = { q: "l" };
+  params = { q: "" };
   url: URL;
   constructor() {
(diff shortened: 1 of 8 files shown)


pranavnarang (approved):
LGTM!

PR #143, opened 2020-11-17
[#124] fix: side bar proportion issue 

### LINKED ISSUE #142: Fix the side bar proportion issue
## Description 
I realized that the current side bar is not extended to full page on a different screen size and currently this behaviour only happen in Chrome browser

## Sub-task
- [x] fixing the size bar component
- [x] ensure that there is no style overwriting

IanSeng:
# Summary 
This pr is to fix the current issue of the side bar not extending to full page on a different screen size. This PR also included removing unnecessary styling and moved duplicated style into global css. 

### Sub-tasks
- [x] fixing the size bar component
- [x] ensure that there is no style overwriting
- [x] manual tested the visual on (Safari & Chrome)
- [x] extract duplicate styles

### Issue
Fixes #142

### DIFF
--- a/public/global.css
+++ b/public/global.css
@@ -36,6 +36,7 @@
 body {
   font-family: var(--main-font), sans-serif;
   margin: 0;
+  padding: 0;
 }
 
 h1, h2, h3, h4, h5, h6 {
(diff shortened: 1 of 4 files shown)


eddieantonio (approved):
LGTM! 

This is the right way to do it!

PR #145, opened 2020-11-18
factor out upload logic

### LINKED ISSUE #144: Factor out upload logic for both landing page and language page

xqiu1:
Fixes #144


eddieantonio (changes requested):
This is awesome! I really like how much simpler this makes both the `Upload` and the `AddSource` compoonent, and the factoring is much better than it was before (no more "download is part of upload"!).

~~...however, I'm wary of what will happen when this is merged with `production`. I think the changes introduced in #139 might interfere with the changes here. Could you please merge `production` into this branch, and make sure everything still works?~~

EDIT: ignore that 😅

eddieantonio (approved):
LGTM!

(lololololololoololololololoololl ignore my comment, there is straight up a bug in production llloplololololololololo)

Issue #150, opened 2020-11-20
Enhance the current testing for autocomplete component with Keyman api data


IanSeng:
# Background 
- The test is not able to mock the indexedDB to do a through e2e testing for the component with the Keyman api data
- The testing code is currently being commented out

### Possible solution
- [ ] Clear out existing data IndexedDB and input a mock data to isolate the test with actual data.

eddieantonio:
I don't understand the context of this issue anymore, so I'm marking it as `wontfix`. Can be reopened in the future.

PR #152, opened 2020-11-20
[#33] feat: Add drop down to associate the current dictionary with an existing FirstVoices/Keyman keyboard.

### LINKED ISSUE #33: Add drop down to associate the current dictionary with an existing FirstVoices/Keyman keyboard.

IanSeng:
# Summary
~~**Note**: This PR should not be merge before  [#132](https://github.com/eddieantonio/predictive-text-studio/issues/132) goes into production branch~~ 
**New Note**: This pr has been rebased with production branch.
This pr is to leverage on the autocomplete component create in [#132](https://github.com/eddieantonio/predictive-text-studio/issues/132) to allow user to select the available Keyman keyboard. 

## Subtask
- [x] Retrieve Keyman keyboard data from worker
- [x] Remove the TODO code that has introduced in [#132](https://github.com/eddieantonio/predictive-text-studio/issues/132)
- [x] Replace the current input field for language with the new autocomplete component
 
## Issue with testing 😭
After discussion with @eddieantonio, we have agreed on creating another task for investigating and resolving the issue of testing. More background: [#150](https://github.com/eddieantonio/predictive-text-studio/issues/150)

## Resolve 
#33


eddieantonio:
An animation of autocomplete working with a few languages


✨  THIS IS SO COOL ✨

src/app/components/AutoComplete.svelte
```diff
@@ -24,13 +36,17 @@
   // Index of focus element
   let index = -1;
 
+  onMount(async () => {
+    results = await worker.getDataFromStorage();
+  });
+
   // Input to search
-  function onChange(event: Event) {
+  async function onChange(event: any) {
```
eddieantonio:
```suggestion
  function onChange(event: any) {
```

Also, why did the typing change from `Event` to `any`?

src/app/components/AutoComplete.svelte
```diff
@@ -1,20 +1,32 @@
...
     bcp47Tag: string;
+    /**
+     * Keyboard Language of the bcp47Tag
+     */
     language: string;
+    /**
+     * Timestamp of last updated time
+     */
+    timestamp: Date;
   }
```
eddieantonio: Can this interface be placed in `src/common/types.ts`? That way, it's not duplicated all over the place. Note that TypeScript has "structural subtyping" so an interface doesn't need to _explicilty_ `extend` a different interface to be compatible.

eddieantonio (changes requested):
I'm really excited for this to drop! However, I think this could use a little bit more refinement. See my comments!

src/app/components/AutoComplete.svelte
IanSeng: I moved the interface to `src/common/types.ts` but Svelte is complaining that the interface is not being exported from the file although I have put an export there 😭. Is this because the component file is being bundled before the worker in which it doesn't not recognize the interface is in the file?

IanSeng (commented):
> I'm really excited for this to drop! However, I think this could use a little bit more refinement. See my comments!

I have addressed the the comments accordingly🎉. After playing with this component, I realized that it is missing a spinner 
(or loading text) when the indexedDB is not ready and show no result found if the input a valid language. I will further refine this autocomplete component once I have done the high priority task 🎉

src/app/components/AutoComplete.svelte
IanSeng: I should have rebased production before requesting for a review, sorry for the confusion. 🤓

src/app/components/AutoComplete.svelte
```diff
@@ -24,13 +36,18 @@
   // Index of focus element
   let index = -1;
 
+  onMount(async () => {
+    results = await worker.getDataFromStorage();
+  });
+
   // Input to search
-  function onChange(event: Event) {
+  async function onChange(event: Event) {
```
eddieantonio: lol why does this keep getting `async`?

```suggestion
 function onChange(event: Event) {
```

src/app/components/AutoComplete.svelte
eddieantonio:
```suggestion
  import type { KeyboardDataWithTime } from "@common/types";
```

We ran into this issue before ;_; you need to `import type` or else svelte-preprocess will try to actually import some sort of concrete object from `@common/types`. See:

https://github.com/sveltejs/svelte-preprocess/blob/main/docs/preprocessing.md#typescript---limitations

eddieantonio (approved):
lol, I just commited my suggestions. LGTM!

IanSeng:
> lol, I just commited my suggestions. LGTM!

OMG hahahaha! thank you so much!

PR #154, opened 2020-11-20
feat(cypress): cy.data() command

eddieantonio:
Creates a new Cypress command: `cy.data(dataCy)`. This is a simple shortcut for `cy.get("[data-cy=NAME]")`. I've also changed all existing instances of `cy.get("[data-cy=NAME]")` to `cy.data("NAME")`.

Spun off from [a comment](https://github.com/eddieantonio/predictive-text-studio/pull/136/files#r528016522)
in #136.


xqiu1 (approved):
cy.data() makes sense to me!

PR #155, opened 2020-11-22
Add support for manual entry

### LINKED ISSUE #37: Add support to enter word and count manually

xqiu1:
Fixes #37


src/app/components/ManualEntry.svelte
```diff
@@ -1,9 +1,57 @@
 <script lang="ts">
-  /**
-   * Handles the click when the delete button is pressed
-   * TODO
-   */
-  const deleteRow = (): void => {};
+  import Button from "./Button.svelte";
+  import worker from "../spawn-worker";
+
+  interface rowDataObj {
```
eddieantonio: Interface names should be in PascalCase and should avoid the words `obj` and `interface` in the name. How about `DictionaryEntry`?

```suggestion
  interface DictionaryEntry {
```

(See: https://github.com/airbnb/javascript#naming--PascalCase)

src/app/components/ManualEntry.svelte
```diff
@@ -1,9 +1,57 @@
...
+    word: string;
+    count: number | undefined;
```
eddieantonio:
```suggestion
    count?: number;
```

src/app/components/ManualEntry.svelte
```diff
@@ -1,9 +1,57 @@
...
+    },
+  ];
+  $: validInput = rowDataFromManualEntry.every((element) =>
+    validateRowData(element)
+  );
+  $: btnColor = validInput ? "blue" : "grey";
+
+  const validateRowData = (rowData: rowDataObj): boolean => {
+    const word = rowData.word;
+    return word !== "" && word !== undefined && word !== null;
```
eddieantonio: Please add a check that word is not just whitespace! e.g., `"  \t \n"` is not a word in any language!

src/app/components/ManualEntry.svelte
```diff
@@ -42,39 +83,46 @@
...
+  table tr:last-of-type td:first-of-type {
     border-bottom-left-radius: 10px;
   }
-  tr:last-of-type td:last-of-type {
+  table tr:last-of-type td:last-of-type {
     border-bottom-right-radius: 10px;
   }
 
-  .btn--inline {
+  table .btn--inline {
```
eddieantonio: Svelte has "scoped styles" so these styles only apply to elements in _this_ `.svelte` component. Why did you add the extra `table ` to all these selectors? Especially since `td`, `th` and `tr` _must_ be children of a `<table>` element.

src/app/components/ManualEntry.svelte
```diff
@@ -42,39 +83,46 @@
...
+  table tr {
     border-bottom: 1pt solid var(--lite-white);
   }
 
-  td {
+  table td {
     vertical-align: middle;
     padding: 0.625rem 0 0.625rem 1.25rem;
     color: var(--gray-dark);
     text-align: center;
```
eddieantonio: I think center alignment should be avoided. [See this answer on UX stack exchange](https://ux.stackexchange.com/a/113594). The correct alignment depends on the nature of the content, but I think for words it should be left aligned for LTR langauges (e.g., English, Chinese) and right-aligned for RTL languages (e.g., Arabic Hebrew). You can use the `text-align: start` for this:

```suggestion
    text-align: start;
```

Arabic numbers (i.e., 1234567890) should _always_ be right-aligned.

src/app/components/ManualEntry.svelte
```diff
@@ -42,39 +83,46 @@
...
     display: inline;
     padding: 0;
     border: 0;
     background-color: transparent;
     font-size: 1em;
   }
 
-  td > input {
+  table td > input {
     height: 20px;
```
eddieantonio: Use `rem` units for this measure, so that it scales with the containing font-size

src/app/components/ManualEntry.svelte
```diff
@@ -84,50 +132,87 @@
     font: 1em var(--main-font);
   }
 
-  input:focus {
+  table input:focus {
     background-color: var(--lite-white);
   }
+
+  table p {
+    display: inline;
```
eddieantonio: I strongly advice against setting `p` to have `display: inline`. `<p>` is semantically a block-level element, so the HTML spec actually treats it as a "block" even if it appears inline.

src/app/components/ManualEntry.svelte
```diff
@@ -84,50 +132,87 @@
...
+    {#each rowDataFromManualEntry as row, i (i)}
+      <tr>
+        <td>
+          <input
+            type="text"
+            bind:value={row.word}
+            data-cy="manual-entry-input-word" />
+          {#if !validateRowData(row)}
+            <p>*</p>
+          {/if}
```
eddieantonio: Some of the validation can be done automatically by the browser. If you want to make sure the field is filled, use [the `required` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/required). No JavaScript needed!

```suggestion
          <input
            type="text"
            bind:value={row.word}
            required
            data-cy="manual-entry-input-word" />
```

I would not keep the \*, but if you really want it, wrap it in a `<span>`; not a `<p>`

src/app/components/ManualEntry.svelte
```diff
@@ -84,50 +132,87 @@
...
-          data-cy="manual-entry-input-count" />
-      </td>
-      <td>
+      <td colspan="3">
         <button
           class="btn--inline"
-          on:click={deleteRow}
-          data-cy="manual-entry-delete">Delete</button>
+          on:click={addNewRow}
+          data-cy="munual-entry-add-row">Add Row</button>
```
eddieantonio: Typo:

```suggestion
          data-cy="manual-entry-add-row">Add Row</button>
```

src/app/components/ManualEntry.svelte
```diff
@@ -1,9 +1,57 @@
...
+    const word = rowData.word;
+    return word !== "" && word !== undefined && word !== null;
+  };
+
+  const deleteRow = (i: number): void => {
+    const part = rowDataFromManualEntry.splice(i, 1);
+    rowDataFromManualEntry = rowDataFromManualEntry;
+  };
+
+  const addNewRow = (): void => {
```
eddieantonio, line 49: I really like the fact that this is called `addNewRow` and not like... `handleClick`. It makes sense EXACTLY what this will do when clicked!

src/app/components/ManualEntry.svelte
```diff
@@ -84,50 +132,87 @@
...
-                <button class="btn--inline" on:click={deleteRow}>Delete</button>
-            </td>
-        </tr>
   </table>
+
+  <div class="save-zone">
+    <Button
+      color={btnColor}
+      isOutlined
+      disabled={!validInput}
```
eddieantonio: I do not like this solution (having a manual save button), as I think an auto-save option is better. However, I think autosave is enough work that it's a new PR 😅 

The problem with a button that gets disabled on invalid input is that there's no written feedback as to what exactly is wrong with the form. "Why can't I save"? The `required` attribute of `<input>` would help with this. Note: you may want to wrap the entire table in `<form>` element, and then make the underlying button here be a `<button type="submit">`. This will enable the browser to automatically validate the `required` attribute AND tell you which words are missing!

src/common/predictive-text-studio-worker.ts
```diff
@@ -12,13 +12,21 @@ export interface PredictiveTextStudioWorker {
...
   addDictionarySourceToProject(name: string, contents: File): Promise<number>;
 
+  /**
+   * Store the manual entry data into the database
+   * @param tableData Manual entry data
+   */
+  addManualEntryDictionaryToProject(tableData: {
+    name: string;
+    data: { word: string; count: number | undefined }[];
+  }): Promise<number>;
```
eddieantonio: Could you please extract an interface for this type?

src/worker/read-wordlist.ts
```diff
@@ -45,3 +45,15 @@ function asNonNegativeInteger(x: unknown): number {
 
   return n;
 }
+
+export async function readManualEntryData(
+  contents: { word: string; count: number | undefined }[]
+): Promise<WordList> {
```
eddieantonio: Only functions that actually use promises or other the `await` syntax should be marked as `async`:

```suggestion
export function readManualEntryData(
  contents: { word: string; count: number | undefined }[]
): WordList {
```

src/worker/read-wordlist.ts
```diff
@@ -45,3 +45,15 @@ function asNonNegativeInteger(x: unknown): number {
...
+  const wordlist: WordList = [];
+  contents.forEach((row) => {
+    const word = row.word;
+    const count = asNonNegativeInteger(row.count || 0);
+    wordlist.push([word, count]);
+  });
+  return wordlist;
```
eddieantonio: This is a good candidate for `Array.prototype.map()`:

```suggestion
  const wordlist: WordList = contents.map((row) => {
    const word = row.word;
    const count = asNonNegativeInteger(row.count || 0);
    return [word, count];
  });
  return wordlist;
```

eddieantonio (changes requested):
I have a LOT of nit picky feedback. Please feel free to DM me if you have any questions!

src/app/components/ManualEntry.svelte
xqiu1: Sure, it makes more sense

src/app/components/ManualEntry.svelte
```diff
@@ -1,9 +1,80 @@
...
+      name: "action",
+    },
+  ];
+  $: rowDataFromManualEntry = tableData.data;
+  $: validDictionary =
+    validateRowsData(rowDataFromManualEntry) &&
+    validateTableName(tableData.name);
+  $: if (validDictionary) {
+    worker.addManualEntryDictionaryToProject(tableData);
+  }
```
xqiu1: I add the basic support for autosaving dictionary.

src/app/components/Button.svelte
```diff
@@ -28,6 +29,7 @@
     class="button button--{color} button--{size}"
     class:button--shadow={hasDropShadow}
     class:button--outline={isOutlined}
+    {type}
```
eddieantonio, line 32: This is a great change! I found out recently that [the default type of a button is `type="submit"`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-type) which... is not always helpful 😅

src/app/components/ManualEntry.svelte
```diff
@@ -1,9 +1,80 @@
...
+    },
+    {
+      label: "Count",
+      name: "count",
+    },
+    {
+      label: "Action",
+      name: "action",
+    },
+  ];
```
eddieantonio: Honestly, I'd rather these columns be hardcoded in the HTML below for the time being rather than using this array. I'm not sure what benefit this array brings.

src/app/components/ManualEntry.svelte
```diff
@@ -1,9 +1,80 @@
...
+
+  const validateRowsData = (rowsData: DictionaryEntry[]): boolean => {
+    return rowsData.every((rowData) => {
+      const word = rowData.word;
+      return (
+        word !== "" &&
+        word !== undefined &&
+        word !== null &&
+        !word.match(/^([\s\t\r\n]*)$/)
+      );
```
eddieantonio: I feel like there's some code duplication that could be factored out here! In another PR though!

src/app/components/ManualEntry.svelte
```diff
@@ -1,9 +1,80 @@
...
+
+  const saveTableData = async () => {
+    const validDictionary: boolean =
+      validateRowsData(rowDataFromManualEntry) &&
+      validateTableName(tableData.name);
+    if (validDictionary) {
+      const numOfWordStored = await worker.addManualEntryDictionaryToProject(
+        tableData
+      );
+    }
```
eddieantonio: There's some duplicated code here with the reactive statementsthat handle autosave (`$: if (validDictionary) ...`). Can you reduce the duplication here?

src/app/components/ManualEntry.svelte
```diff
@@ -87,47 +159,78 @@
...
+    display: flex;
+    flex-direction: row;
+    align-items: center;
+    justify-content: center;
+    padding: 1%;
+  }
 </style>
 
-<div class="language__sources-manual-entry">
+<form class="language__sources-manual-entry" on:submit|preventDefault>
```
eddieantonio, line 155: 👍🏼

src/app/components/ManualEntry.svelte
```diff
@@ -87,47 +159,78 @@
...
-  <table>
+  <table id="manual-entry-table">
     <thead>
-      <th>Word</th>
-      <th>Count</th>
-      <th>Action</th>
+      {#each columns as column (column.name)}
+        <th>{column.label}</th>
+      {/each}
+      <th />
```
eddieantonio: Was this left here by mistake?

```suggestion
```

src/app/components/ManualEntry.svelte
```diff
@@ -87,47 +159,78 @@
...
 
-  <table>
+  <table id="manual-entry-table">
     <thead>
-      <th>Word</th>
-      <th>Count</th>
-      <th>Action</th>
+      {#each columns as column (column.name)}
+        <th>{column.label}</th>
+      {/each}
```
eddieantonio: This is what I meant. I think it's best to just hard code the headers here.
```suggestion
      <th> Word </th>
      <th> Count </th>
      <th> Action </th>
```

src/app/components/ManualEntry.svelte
```diff
@@ -87,47 +159,78 @@
...
+            type="number"
+            min="0"
+            placeholder="Optional"
+            bind:value={row.count}
+            data-cy="manual-entry-input-count" />
+        </td>
+        <td>
+          <button
+            class="btn--inline"
+            on:click={() => deleteRow(i)}
```
eddieantonio, line 192: very nice!

src/app/components/ManualEntry.svelte
```diff
@@ -87,47 +159,78 @@
...
-                <button class="btn--inline" on:click={deleteRow}>Delete</button>
-            </td>
-        </tr>
   </table>
-</div>
+
+  <div class="save-zone">
+    <Button
+      type={'submit'}
+      color={'blue'}
```
eddieantonio:
```suggestion
      type="submit"
      color="blue"
```

src/common/manual-entry-file.ts
```diff
@@ -0,0 +1,4 @@
+export interface DictionaryEntry {
```
eddieantonio: I like the new interface, but could you please put it in in `src/common/types.ts`? I don't want there to be an explosion of files. Not every interface needs its own source file!

Also, please add JSDoc comments!

src/worker/predictive-text-studio-worker-impl.ts
```diff
@@ -87,6 +88,16 @@ export class PredictiveTextStudioWorkerImpl
     return wordlist.length;
   }
 
+  async addManualEntryDictionaryToProject(tableData: {
+    name: string;
+    data: DictionaryEntry[];
+  }): Promise<number> {
+    const dictionaryName = tableData.name;
+    const wordlist = await readManualEntryData(tableData.data);
```
```suggestion
    const wordlist = readManualEntryData(tableData.data);
```

src/worker/read-wordlist.ts
```diff
@@ -45,3 +45,12 @@ function asNonNegativeInteger(x: unknown): number {
 
   return n;
 }
+
+export function readManualEntryData(contents: DictionaryEntry[]): WordList {
```
eddieantonio, line 49: 👍🏼

eddieantonio (changes requested):
Looking better! There are still a few things to polish up!

src/app/components/ManualEntry.svelte
xqiu1: Sure, I will try to factor out the duplication 😅

eddieantonio (approved):
LGTM!

PR #156, opened 2020-11-23
Fix/accessibility

### LINKED ISSUE #148: feat: enhance accessibility all pages
enhance accessibility of all pages and fix the potential accessibility issue
Suggested to use the following tools 
- Lighthouse or Axe

IanSeng:
# Summary 
This pr is to enhance and fix the existing accessibility issues.

## Improvement
**Lighthouse performance report** 
Landing page: improved from `89% -> 100%`
Language page: improved from` 81% -> 100% `

**Axe report** 
Landing page:fixed almost all issues found by axe, but omitted the fix for changing the color the header text such as "Predictive Text Studio", "designed for Keyman" and "Add prediction and autocorrect to your language", since they look decent. 
Language page: addressed all issue reported by axe

## Changes & rationals 
- enhanced the input component logic to incorporate `label` for screen reader users. 
**Tradeoff**: Instead of using the `if and else` statement in the component to hide label for a different design input field (`Step 1: Enter your languages`) in the landing page, a `global` and `!important` are used to override the pre-defined component style. This is ensure better reusability of the input component. 
- modified the primary blue to a darker color to have a better contrast of blue with white back ground. 
- modified the on-hover color to a even darker color for the hover effect 
- modified the cue-text color to a darker color for a better visual contrast  
- added cyData as a properties for input-filed component for easier cypress testing

## Subtasks
- [x] run axe accessibility test 
- [x] run lighthouse accessibility test
- [x] fix existing accessibility issue 
- [x] fix testing
- [x] enhance the existing input component 

## Issue 
Resolves #156

### DIFF
--- a/public/global.css
+++ b/public/global.css
@@ -8,12 +8,14 @@
   --orange-light: #fbc88f;
   --gray: #888888;
   --gray-hover: #c5c5c5;
+  --gray-disabled: #949494;
   --gray-light: #d6d6d6;
   --gray-highlight: #e0e0e0;
   --gray-dark: #454545;
   --lite-white: #f1f1f1;
   --white: #ffffff;
-  --primary-blue: #0099ff;
+  --hello: #0071bc;
+  --primary-blue: #0077cc;
   --dark-blue: #0876c0;
 
   --default-text-color: #000;
--- a/public/global.css
+++ b/public/global.css
@@ -30,7 +30,7 @@
   --xxs: 0.75em;
   --xs: 0.875em;
   --s: 1em;
-  --m: 1.125em; 
+  --m: 1.125em;
   --l: 1.25em;
   --xl: 1.5em;
   --xxl: 2.5em;
@@ -51,11 +51,18 @@ body {
   padding: 0;
 }
 
-h1, h2, h3, h4, h5, h6 {
+h1,
+h2,
+h3,
+h4,
+h5,
+h6 {
   font-family: var(--main-font), sans-serif;
 }
 
-pre, code, samp {
+pre,
+code,
+samp {
   font-family: var(--mono-font), monospace;
 }
 
@@ -66,10 +73,10 @@ pre, code, samp {
  */
 .button {
   display: inline-block;
-  margin: 0 .625em;
-  padding: .625rem 1.25rem; /* button--medium size */
+  margin: 0 0.625em;
+  padding: 0.625rem 1.25rem; /* button--medium size */
   border: 1px solid transparent;
-  border-radius: .625rem;
+  border-radius: 0.625rem;
 
   font: bold 1em var(--main-font), sans-serif;
   text-transform: uppercase;
@@ -84,7 +91,7 @@ pre, code, samp {
 
 /* make <a> .button have a focus ring like <button> elements. */
 a.button:focus {
-  outline: .5rem auto var(--primary-blue);
+  outline: 0.5rem auto var(--primary-blue);
   outline-color: -webkit-focus-ring-color;
 }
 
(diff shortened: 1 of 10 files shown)


public/global.css
```diff
@@ -54,10 +63,10 @@ pre, code, samp {
  */
 .button {
   display: inline-block;
-  margin: 0 .625em;
-  padding: .625rem 1.25rem; /* button--medium size */
+  margin: 0 0.625em;
+  padding: 0.625rem 1.25rem; /* button--medium size */
   border: 1px solid transparent;
-  border-radius: .625rem;
+  border-radius: 0.625rem;
```
eddieantonio: Did your editor's linter do this?

src/app/components/BCP47Tag.svelte
```diff
@@ -14,8 +14,19 @@
   }
 </script>
 
+<style>
+  :global().input-field__label-bcp {
```
eddieantonio: I am SUPER CONFUSED about this selector.

src/app/components/BCP47Tag.svelte
```diff
@@ -14,8 +14,19 @@
...
+    font: var(--main-font);
+    font-size: var(--s) !important;
+    font-weight: normal !important;
```
eddieantonio: pwease do not resort to `!important`. `!important` is a measure of last resort, because it actually overrides the user settings, and the users' settings are what gives us the correct font-size. Why did you use `!important` here, and can that be solved with a more specific selector?

public/global.css
```diff
@@ -7,13 +7,15 @@
   --orange: #f18407;
   --orange-light: #fbc88f;
   --gray: #888888;
+  --gray-secondary-text: #515154;
   --gray-hover: #c5c5c5;
+  --gray-disabled: #949494;
```
eddieantonio, line 12: I really like these semantic names! 🤩

eddieantonio (changes requested):
I have some concerns around the use of `!important` and confusing selectors before I can approve this, but I do like the increased Lighthouse and Axe scores! And those semantic colour names... [chef's kiss]

public/global.css
IanSeng: lol yeah my Prettier plug-in added all of these 0. Should I remove these zeroes

src/app/components/BCP47Tag.svelte
IanSeng: the `.input-field__label-bcp` should be inside the global()

src/app/components/BCP47Tag.svelte
IanSeng: correct if I am wrong, Svelte first compile the page and then the components within the page. Since input field on the landing page have different design, I would have to use important to overwrite the input field style. I personally don't think this is a good solution, I will try a different implementation

IanSeng:
> I have some concerns around the use of `!important` and confusing selectors before I can approve this, but I do like the increased Lighthouse and Axe scores! And those semantic colour names... [chef's kiss]

- I think those are very critical and useful comments and I have addressed them accordingly. 👍🎉

src/app/components/BCP47Tag.svelte
IanSeng: I totally agree on you and `!important` is definitely not a best practice. Rather than just overriding the style, I made some changes to introduce the size attribute for the input component.

public/global.css
eddieantonio: Yes please! Remove all the zeros haha!

eddieantonio (approved):
LGTM! Much improved!

IanSeng:
> LGTM! Much improved!

Addressed all comments and merged into production branch 🎉🎉🎉

PR #157, opened 2020-11-24
[#158] feat: disable button on landing page

### LINKED ISSUE #158: Fix: disable button on landing page
# Background 
Currently the link shows a disable cursor on hover but user is still able to click on the link to download an unexpected file. 

## Fix 
- [x] Disable the click functionally to avoid user from downloading this unexpected file by accident

IanSeng:
# Summary 
Discovered a bug of the current disable issue when I am working on my task so I pull in this quick pr to fix it. This pr is to fix the **disabling bug** for the **download link** on the landing page. Instead of just showing a disabled cursor, now it will also disable user to click on the link and show disable cursor at the same time. 

# Issue 
#158

### DIFF
--- a/src/app/components/DownloadKMP.svelte
+++ b/src/app/components/DownloadKMP.svelte
@@ -3,25 +3,32 @@
 </script>
 
 <style>
-  .download-link {
+  .download-component--disabled {
+    cursor: not-allowed;
+  }
+  .download-component__link {
     display: block;
     margin-top: 1.5em;
   }
 
-  .download-link--disabled {
+  .download-component__link--disabled {
     color: var(--gray-medium-dark);
-    cursor: not-allowed;
     opacity: 25%;
+    pointer-events: none;
     text-decoration: none;
   }
 </style>
 
-<a
-  href={downloadURL ? downloadURL : '#'}
-  download="Example.kmp"
-  class="download-link"
-  class:download-link--disabled={downloadURL == ''}
-  data-cy="download-kmp"
-  data-download-state={downloadURL ? 'ready' : 'disabled'}>
-  Download KMP Package
-</a>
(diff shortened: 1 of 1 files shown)


src/app/components/DownloadKMP.svelte
```diff
@@ -3,25 +3,32 @@
...
+  <a
+    href={downloadURL ? downloadURL : '#'}
+    download="Example.kmp"
+    class="download-component__link"
+    class:download-component__link--disabled={downloadURL == ''}
+    data-cy="download-kmp"
+    data-download-state={downloadURL ? 'ready' : 'disabled'}>
+    Download KMP Package
+  </a>
+</div>
```
eddieantonio, line 34: Is the wrapper `<div>` necessary? Why?

IanSeng: Yup, the wrapper `<div>` is to inherit the style of `cursor: not-allowed;` while the `<a>` will only responsible for `pointer-events: none`. Since if `cursor: not-allowed;` & `pointer-events: none` exist together as a style for a single tag, only one of them will be selected, depending on the order. This was the reason why before the fix it will only show disabled cursor but the user is still able to click on the `<a>` tag.  

The behaviour now will be the wrapper class `<div>` giving the style of `cursor: not-allowed;` and the `<a>` tag being refrained from clicking by `pointer-events: none;`

eddieantonio: Oh wow! Yeah, if the `<a>` gets `pointer-events: none` it can't get... pointer events... including the event that changes the cursor!

eddieantonio (approved):
LGTM!

I learned something new abount `pointer-events: none`!

PR #166, opened 2020-11-30
[#163] feat: Remove "Projects" sidebar

### LINKED ISSUE #163: Remove "Projects" sidebar
We have a sidebar that is supposed to allow the user to select between multiple projects. However, we only support one project at the moment! We should get rid of the sidebar while this feature doesn't exist, or at least hide the buttons!

Get rid of this thing (for now):

The sidebar

IanSeng:
# Summary 
This pr is to remove side bar component from the language page. I am not sure if we might need this component in the future, hence, I did not delete the `Sidebar.svelte` component from the codebase. 

### Issue
Fixes #163

### DIFF
--- a/src/app/pages/Languages.svelte
+++ b/src/app/pages/Languages.svelte
@@ -157,10 +157,6 @@
  -->
 <main>
   <div class="languages">
-    <div class="languages__sidebar">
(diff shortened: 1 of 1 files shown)


xqiu1 (approved):
LGTM

PR #174, opened 2020-12-03
Save KMP in the app

### LINKED ISSUE #172: Make the KMP file exist in the app
Currently the kmp package exists only when we click the download link in the landing page, but now we want it can exist in the app independently even before clicking the link.

xqiu1:
Fixes #172


pranavnarang (approved):
LGTM

src/worker/predictive-text-studio-worker-impl.ts
```diff
@@ -146,4 +151,12 @@ export class PredictiveTextStudioWorkerImpl
     };
     return this.storage.updateProjectData(storedData);
   }
+
+  saveKMPPackage(KMP: ArrayBuffer): Promise<void> {
+    return this.storage.saveCompiledKMPAsArrayBuffer(KMP);
+  }
+
+  async getKMPPackage(): Promise<KMPPackageData> {
```
eddieantonio: I think we should keep`KMPPackageData` as an internal data structure, especially since the only place that consumes it just does `kmp.package` :joy:

```suggestion
  async getKMPPackage(): Promise<ArrayBuffer> {
```

src/worker/storage.ts
```diff
@@ -185,4 +197,30 @@ export default class Storage {
...
+   * Retrieve the compiled KMP package from database
+   */
+  async fetchCompiledKMPFile(): Promise<KMPPackageData> {
+    const KMPFile = await this.db.KMPFileData.where(":id")
+      .equals(PACKAGE_ID)
+      .first();
+    if (KMPFile == undefined) {
+      throw new Error("No KMP file has been compiled");
+    }
+    return KMPFile;
```
```suggestion
    const kmpFile = await this.db.KMPFileData.where(":id")
      .equals(PACKAGE_ID)
      .first();
    if (kmpFile == undefined) {
      throw new Error("No KMP file has been compiled");
    }
    return kmpFile;
```

src/worker/storage.ts
```diff
@@ -185,4 +197,30 @@ export default class Storage {
...
   }
+
+  /**
+   * Save the KMP package once it compiled
+   * @param KMPFile
+   */
+  saveCompiledKMPAsArrayBuffer(KMPFile: ArrayBuffer): Promise<void> {
+    return this.db.transaction("readwrite", this.db.KMPFileData, async () => {
+      await this.db.KMPFileData.put({
+        package: KMPFile,
```
```suggestion
  saveCompiledKMPAsArrayBuffer(kmpFile: ArrayBuffer): Promise<void> {
    return this.db.transaction("readwrite", this.db.KMPFileData, async () => {
      await this.db.KMPFileData.put({
        package: kmpFile,
```

src/common/types.ts
```diff
@@ -38,3 +38,8 @@ export interface KeyboardDataWithTime {
    */
   timestamp: Date;
 }
+
+export interface KMPPackageData {
+  /** The coompiled KMP package */
+  package: ArrayBuffer;
+}
```
eddieantonio: This is unnecessary if `worker.getKMPPackage()` returns an `ArrayBuffer`

```suggestion
```

src/common/predictive-text-studio-worker.ts
```diff
@@ -1,5 +1,9 @@
 import type { RelevantKmpOptions } from "./kmp-json-file";
-import type { DictionaryEntry, KeyboardDataWithTime } from "./types";
+import type {
+  DictionaryEntry,
+  KeyboardDataWithTime,
+  KMPPackageData,
```
eddieantonio: See comments below

```suggestion
```

src/app/pages/LandingPage.svelte
```diff
@@ -7,7 +7,9 @@
   let downloadURL = "";
 
   worker.onPackageCompileSuccess(
-    Comlink.proxy((kmp: ArrayBuffer) => {
+    Comlink.proxy(async () => {
+      const KMPPackage = await worker.getKMPPackage();
+      const kmp = KMPPackage.package;
```
eddieantonio: See suggestions below

```suggestion
      const kmp = await worker.getKMPPackage();
```

src/worker/predictive-text-studio-worker-impl.ts
```diff
@@ -77,8 +81,9 @@ export class PredictiveTextStudioWorkerImpl
         new Error("Cannot find any files in the IndexedDB")
       );
     } else {
-      const kmpFile = await linkStorageToKmp(this.storage);
-      this._emitPackageCompileSuccess(kmpFile);
+      const KMPArrayBuffer = await linkStorageToKmp(this.storage);
+      this.saveKMPPackage(KMPArrayBuffer);
```
```suggestion
      const kmpArrayBuffer = await linkStorageToKmp(this.storage);
      this.saveKMPPackage(kmpArrayBuffer);
```

src/worker/predictive-text-studio-worker-impl.ts
```diff
@@ -77,8 +81,9 @@ export class PredictiveTextStudioWorkerImpl
...
+      this._emitPackageCompileSuccess();
```
eddieantonio, line 82: It doen not matter in this project, but when you are creating a public API, this would be considered a breaking change and there's really nothing stopping you from preserving backwards compatibility. You can definitely both save the `KMP` packlage and send it to the `onPackageCompileSuccess()` callback.

Again, in this project it doesn't matter, but something to consider when making APIs. If you wanted people to use `getKMPPackage()` instead, you could add to the documentation that this parameter is deprecated.

```typescript
      this._emitPackageCompileSuccess(kmp);
```

src/worker/storage-models.ts
```diff
@@ -77,3 +77,9 @@ export interface KeyboardDataWithTime extends KeyboardData {
    */
   timestamp: Date;
 }
+export interface KMPPackageData {
+  /** Autogenerated primary key */
+  id?: number;
+  /** The coompiled KMP package */
```
eddieantonio:
```suggestion
  /** The compiled KMP package */
```

src/common/predictive-text-studio-worker.ts
```diff
@@ -60,4 +64,15 @@ export interface PredictiveTextStudioWorker {
    * Retrieving Keyman keyboard data from the IndexedDB storage
    */
   getDataFromStorage(): Promise<KeyboardDataWithTime[]>;
+
+  /**
+   * Save the KMP package once it compiled
+   * @param KMPFile
+   */
+  saveKMPPackage(KMP: ArrayBuffer): Promise<void>;
+
```
eddieantonio: It seems like nothing outside of `PredictiveTextStudioWorkerImpl` uses this method, so I would not expose it as part of the public API.

```suggestion
```

src/worker/predictive-text-studio-worker-impl.ts
```diff
@@ -146,4 +151,12 @@ export class PredictiveTextStudioWorkerImpl
     };
     return this.storage.updateProjectData(storedData);
   }
+
+  saveKMPPackage(KMP: ArrayBuffer): Promise<void> {
+    return this.storage.saveCompiledKMPAsArrayBuffer(KMP);
+  }
```
eddieantonio: This is a method used exclusively in this class; it can become public later, if needed, but right now, it seems to be an implementation detail.

```suggestion
  private saveKMPPackage(kmp: ArrayBuffer): Promise<void> {
    return this.storage.saveCompiledKMPAsArrayBuffer(kmp);
  }
```

src/common/predictive-text-studio-worker.ts
```diff
@@ -60,4 +64,15 @@ export interface PredictiveTextStudioWorker {
...
+  /**
+   * Retrieve the compiled KMP package from database
+   */
+  getKMPPackage(): Promise<KMPPackageData>;
```
```suggestion
  getKMPPackage(): Promise<ArrayBuffer>;
```

test/test-storage.ts
```diff
@@ -195,3 +195,23 @@ test("delete all Keyman keyboard data with deleteKeyboardData", async (t) => {
     t.is(data.length, 0);
   });
 });
+
+test("store the KMP package to database", async (t) => {
+  const { db, storage } = t.context;
+  // At first, nothing in the DB
+  t.is(await db.KMPFileData.count(), 0);
+  const KMP = new ArrayBuffer(1);
+  await storage.saveCompiledKMPAsArrayBuffer(KMP);
```
```suggestion
  const kmp = new ArrayBuffer(1);
  await storage.saveCompiledKMPAsArrayBuffer(kmp);
```

test/test-storage.ts
```diff
@@ -195,3 +195,23 @@ test("delete all Keyman keyboard data with deleteKeyboardData", async (t) => {
...
+});
+
+test("retrieve the KMP package from the database", async (t) => {
+  const { storage } = t.context;
+  const KMP = new ArrayBuffer(1);
+  await storage.saveCompiledKMPAsArrayBuffer(KMP);
+
+  const KMPRetrieved = await storage.fetchCompiledKMPFile();
+  const packageRetrieved = KMPRetrieved.package;
+  t.assert(packageRetrieved instanceof ArrayBuffer);
```
```suggestion
  const kmp = new ArrayBuffer(1);
  await storage.saveCompiledKMPAsArrayBuffer(kmp);

  const kmpRetrieved = await storage.fetchCompiledKMPFile();
  const packageRetrieved = kmpRetrieved.package;
  t.assert(packageRetrieved instanceof ArrayBuffer);
```

eddieantonio (changes requested):
My biggest request is that the API footprint remains small. Bascially, don't expose more than absolutely needed. The app only needs to `worker.getKMPPackage()` and then it only needs the `ArrayBuffer`; it doesn't need an ID or anything, since that's an implementation detail of Dexie/IndexedDB.

My other nitpick is that the naming rules for acronyms conflicts with the rule that variable names and property names should start with a lowercase letter, whereas class names, interface names, and constant names should start with an uppercase letter. I think the latter rule should take precedence -- if the name starts with a capital, it's ALWAYS a class name, an interface, or a constant name. Therefore, any variable or property that starts with `KMP` should be named `kmp` instead!

src/worker/predictive-text-studio-worker-impl.ts
xqiu1: Thank you for your suggestion!

src/worker/predictive-text-studio-worker-impl.ts
xqiu1: Sure. What I thought before was we can store different types of KMP file, so I create the interface. But using ArrayBuffer is absolutely much simpler.

eddieantonio (approved):
LGTM! 🎉

PR #182, opened 2020-12-09
Fix/set project metadata

### LINKED ISSUE #196: Inconsistent results when updating project metadata
The project metadata currently gets completely messed up when editing multiple fields. There should be regression tests to make sure the project data remains consistent even after modifying multiple different fields.

Did I just create an issue so that I can immediately close it? Maaaaybeeee 😇

eddieantonio:
Project metadata was being updated incorrectly. I did the process here backwards, so I'll create an issue report, add some tests, then set this PR as ready for review.

EDIT: fixes #196

### DIFF
--- a/src/worker/storage.ts
+++ b/src/worker/storage.ts
@@ -140,15 +140,9 @@ export default class Storage {
    */
   updateProjectData(metadata: { [key: string]: string }): Promise<void> {
     return this.db.transaction("readwrite", this.db.projectData, async () => {
-      await this.db.projectData.put({
-        langName: metadata.langName,
-        bcp47Tag: metadata.bcp47Tag,
-        authorName: metadata.authorName,
-        modelID: metadata.modelID,
-        copyright: metadata.copyright,
-        version: metadata.version,
-        id: PACKAGE_ID,
-      });
+      const existingMetadata = await this.db.projectData.get(PACKAGE_ID);
+      const updatedMetadata = Object.assign(existingMetadata, metadata);
+      await this.db.projectData.put(updatedMetadata);
     });
   }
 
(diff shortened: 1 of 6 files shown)


src/worker/storage.ts
```diff
@@ -140,15 +140,14 @@ export default class Storage {
...
-        langName: metadata.langName,
-        bcp47Tag: metadata.bcp47Tag,
-        authorName: metadata.authorName,
-        modelID: metadata.modelID,
-        copyright: metadata.copyright,
-        version: metadata.version,
-        id: PACKAGE_ID,
-      });
+      const existingMetadata:
+        | StoredProjectData
```
ChrisChrisLoLo, line 144: Out of curiosity, what do you call the `|` symbol in this context? I can only find references to it as a bitwise OR operator

ChrisChrisLoLo (approved):
LGTM!

Nothing's jumping out at me. I don't think I'm familiar enough with the codebase to be super stringent with the underlying logic of the program, but I should get more familiar the more I work with it and use it.

src/worker/storage.ts
eddieantonio: I'm not sure if it has a name in this context. According to the TypeScript docs:

> We use the vertical bar (`|`) to separate each type, so `number | string | boolean` is the type of a value that can be a `number`, a `string`, or a `boolean`.
>
> _Source_: https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html

eddieantonio: Type union operator? Or it separates elements of a type union? A sum type operator? I don't really know, haha!

PR #206, opened 2021-01-23
[#170] feat: Add a link to go back to the landing page from the language

### LINKED ISSUE #170: Add a link to go back to the landing page from the language
Currently, the only way to go back to the landing page is by editing the URL.
Add a link to the language page to go back to the landing page (e.g to the Predictive Text Studio header)

yehee:
### Summary
This PR is to create a `Go Back` button on the languages page to allow users to go back to the landing page without having to edit the URL explicitly.

Resolves #170

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,7 +67,6 @@
     "exceljs": "~3.5.0",
     "googleapis": "^61.0.0",
     "jszip": "^3.5.0",
-    "svelte-awesome": "^2.3.0",
     "svelte-routing": "^1.4.2",
     "tslib": "^2.0.1"
   },
--- a/package.json
+++ b/package.json
@@ -67,6 +67,7 @@
     "exceljs": "~3.5.0",
     "googleapis": "^61.0.0",
     "jszip": "^3.5.0",
+    "svelte-icons": "^2.1.0",
     "svelte-routing": "^1.4.2",
     "tslib": "^2.0.1"
   },
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
     "exceljs": "~3.5.0",
     "googleapis": "^61.0.0",
     "jszip": "^3.5.0",
-    "svelte-icons": "^2.1.0",
+    "svelte-awesome": "^2.3.0",
     "svelte-routing": "^1.4.2",
     "tslib": "^2.0.1"
   },
--- a/src/app/pages/Languages.svelte
+++ b/src/app/pages/Languages.svelte
@@ -1,6 +1,5 @@
 <script lang="ts">
   import { Link } from "svelte-routing";
-  import ArrowLeft from "svelte-icons/fa/FaArrowLeft.svelte";
   import IoIosArrowBack from "svelte-icons/io/IoIosArrowBack.svelte";
   import LanguageInfo from "../components/LanguageInfo.svelte";
   import LanguageSources from "../components/LanguageSources.svelte";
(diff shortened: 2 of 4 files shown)


src/app/pages/Languages.svelte
```diff
@@ -124,6 +133,11 @@
 <main>
   <div class="languages">
     <div class="languages__container">
+      <Link to="/">
```
ChrisChrisLoLo: I think these link elements could be why the test runner is complaining, though I'm not sure why it's only complaining on the runner. 

Three potential fixes in my head could be:
- Do as the runner says and try `npm install @types/svelte-routing` if it exists or add a new declaration (.d.ts) file containing `declare module 'svelte-routing';` (ts)
- Use `<a>` tags with `href`s. Svelte seems to play nicely with these
- Something else entirely

Let me know what the fix is! I'm curious about why the runner's failing

package.json
```diff
@@ -67,6 +67,7 @@
     "exceljs": "~3.5.0",
     "googleapis": "^61.0.0",
     "jszip": "^3.5.0",
+    "svelte-icons": "^2.1.0",
```
ChrisChrisLoLo: Good idea 👍

ChrisChrisLoLo (commented):
Changes look good! If you can push a fix for the unit tests, let me know and I can approve

src/app/pages/Languages.svelte
eddieantonio: `yarn add @types/svelte-routing` please!

package.json
eddieantonio: Haha, I'm gonna veto this!

I'd prefer _not_ to include a library in the `app` bundle to add a _single_ icon. I'd have to measure how much larger the app bundle gets by adding this library. This is definitely a user experience vs. developer experience trade-off, and in terms of frontend code, I usually favour user experience, even if it makes things more annoying or complicated for us devs.

In general, adding dependencies to the `worker` bundle is fine, but adding things to the `app` I'm a lot more cautious about.

src/app/pages/Languages.svelte
eddieantonio: I'm really confused because we're using `svelte-routing` elsewhere, and TypeScript does not complain about it.

A workaround is using `<a href="/">` since this _particular_ `<Link>` tag is semantically identical to that.

package.json
eddieantonio: Okay, I did the calculations and it only adds 802 bytes, so it's fine haha! 😅

src/app/pages/Languages.svelte
```diff
@@ -124,6 +133,11 @@
...
+        <span class="button button--grey button--outline mt-xxl">
+          <div class="icon"><IoIosArrowBack /></div>Go back
+        </span>
+      </Link>
```
eddieantonio: This markup should be simplified.

```suggestion
      <a href="/" class="button button--grey button--outline mt-xxl">
          <IoIosArrowBack /> Go back
      </a>
```

For now, do not use the `<Link>` element, because `svelte-check`/TypeScript is having a coniption about it, and this particular usage of the `<Link>` component does not use routes; it is a link to "/" which can be done with an `<a>` element.

As for the size of the icon, I would customize its `viewport` to be the desired size as well as tweak its vertical alignment. The `<IoIosArrowBack />`becomes a `<svg>` in the DOM, so you can target that in selectors!

eddieantonio (changes requested):
Let's pair on this and figure it out!

eddieantonio (approved):
LGTM!

PR #217, opened 2021-01-23
Feat/remove headers

### LINKED ISSUE #199: Remove headers from uploaded wordlists
Remove headers such as "word" and "count" headers in the first row of all uploaded wordlists (either Google Sheets or Excel).

This should be done automatically.

ChrisChrisLoLo:
This feature removes excel rows from being converted into the wordlist if:
- It is a header row (contains the word 'count' under the count column)
- It is a commented out row (the word column is equal to '#')

Partially Resolves #199. Also removes commented out rows.
The logic for the commented rows was added due to a potential keyman workflow found here, https://help.keyman.com/developer/13.0/guides/lexical-models/tutorial/step-3,
specifically in this image: https://help.keyman.com/cdn/deploy/img/developer/120/lm/template-tsv.0f5b50d31126fd5c13f03567fd75d37f.png

EDIT: There currently are some blockers with the Google Sheets API. This feature does not work with the API, and work needs to be done in order to test the API locally, namely #221. I would prefer to get this merged in and create a separate issue for the Google Sheets API than to rather than let this PR go stale while I work on unblocking the API


ChrisChrisLoLo:
Partially blocked by #221. I cannot verify that this feature will work for the google sheets. Would it be possible to break out a new issue for the API and review this in the meantime?

eddieantonio (approved):
LGTM!

PR #225, opened 2021-01-23
Feat/tsv wordlist

### LINKED ISSUE #200: Import wordlist from TSV
Allow import using the TSV format given on this page: https://help.keyman.com/developer/current-version/reference/file-types/tsv

Although users are encouraged to use something they're familiar with, such Google Sheets of Microsoft Excel, they should be able to use this option as well!

zprobs:
Allows users to upload a TSV file to the Input. 

Note: to see the desired behaviour drop a valid .tsv file in the input and check Application/IndexedDB/files/wordlist in dev tools

Resolves #200


src/worker/predictive-text-studio-worker-impl.ts
```diff
@@ -82,7 +82,13 @@ export class PredictiveTextStudioWorkerImpl
     name: string,
     contents: File
   ): Promise<number> {
-    const wordlist = await readExcel(await contents.arrayBuffer());
+    let wordlist: WordList = [];
+    if (/\.(tsv)$/i.test(name)) {
+      wordlist = await readTSV(contents);
+    } else {
```
ChrisChrisLoLo: Should we add logic here to throw an error if the file is neither a .tsv file nor an Excel file? Not entirely sure what Excel files are supported currently

src/worker/read-wordlist.ts
```diff
@@ -38,6 +38,33 @@ export async function readExcel(
...
+        const rows = event.target.result.split("\n");
+        for (let i = 1; i < rows.length; i++) {
+          const row = rows[i].split("\t");
+          if (row.length !== 2) {
+            continue;
+          }
+          const word = row[0] || "";
+          const count = asNonNegativeInteger(row[1] || 1);
+          wordlist.push([word, count]);
+          resolve(wordlist);
```
ChrisChrisLoLo: Do we want to resolve our `wordlist` within the `for` loop?

src/worker/read-wordlist.ts
```diff
@@ -38,6 +38,33 @@ export async function readExcel(
...
+export async function readTSV(TSVFile: File): Promise<WordList> {
+  return new Promise((resolve, reject) => {
+    const wordlist: WordList = [];
+
+    const reader = new FileReader();
+    reader.addEventListener("load", (event: ProgressEvent<FileReader>) => {
+      if (event.target && typeof event.target.result === "string") {
+        // get file contents
+        const rows = event.target.result.split("\n");
+        for (let i = 1; i < rows.length; i++) {
```
ChrisChrisLoLo: Is there a reason for starting the index at 1?

src/worker/predictive-text-studio-worker-impl.ts
```diff
@@ -82,7 +82,13 @@ export class PredictiveTextStudioWorkerImpl
...
+      wordlist = await readExcel(await contents.arrayBuffer());
```
ChrisChrisLoLo, line 101: I wrote some business logic about what kind of rows should be ignored (header rows, commented out rows, etc) for excel files. Is there a way to get both files to share the same kind of business rules functions (i.e. https://github.com/eddieantonio/predictive-text-studio/blob/c0db7a23e0aedb27300caf2821c18d93df5b5ca8/src/worker/read-wordlist.ts#L45)? We could potentially break this out into a separate issue if you'd like.

ChrisChrisLoLo (changes requested):
Thanks for doing this feature! It'll make using the program easier.

I've done my first pass through, though I have not yet tested it.

Could you respond to my feedback, as well as create an automated test for this feature? The test could look something like this: https://github.com/eddieantonio/predictive-text-studio/blob/production/test/test-read-wordlist.ts

Feel free to comment on my points if you disagree, though I might be firm on getting at least one test committed

src/worker/read-wordlist.ts
```diff
@@ -38,6 +38,33 @@ export async function readExcel(
...
+          const row = rows[i].split("\t");
+          if (row.length !== 2) {
```
ChrisChrisLoLo: In `readExcel()` we will still read a line even if it has more than two columns. Could you compare the two functions to ensure that the two functions behave similarly? 

Big bonus points if you find a way to extract that kind of logic out as a common function

zprobs:
@ChrisChrisLoLo All your suggestions have been implemented. Let me know if you spot anything else.

src/worker/read-wordlist.ts
```diff
@@ -38,6 +38,40 @@ export async function readExcel(
...
+  for (let i = 0; i < rows.length; i++) {
+    // Remove BOM from beginning of the string
+    rows[i] = rows[i].replace(/^\uFEFF/, "");
+
+    if (rows[i].startsWith("#") || rows[i] === "") {
+      continue; // skip comments and empty lines
+    }
+
+    // eslint-disable-next-line prefer-const
+    let [word, countText, _comment] = rows[i].split("\t");
```
eddieantonio: Why is there this eslint error?

Honestly, I disagree with the eslint rule of prefer `const` over `let` because `const` is misleading, haha! But regarless, why not just change it to `const`?

Also, (I think I wrote the original code here, haha!) but to sush the warning:

```suggestion
    const [word, countText, /* comment */] = rows[i].split("\t");
```

src/app/components/Upload.svelte
```diff
@@ -102,6 +113,10 @@
     color: var(--blue);
     padding-bottom: 1em;
   }
+  .error {
+    background-color: #f8d7db;
+    color: #400000;
```
ChrisChrisLoLo: Something we can do to maintain consistency is use the css variables we have in `public/global.css`.
We should make the css something along the lines of:
```css
.error {
    background-color: var(--red);
    color: var(--white);
}
```

Or something similar with your color of choice

src/app/components/Upload.svelte
```diff
@@ -38,8 +39,13 @@
...
     }
-    for (let file of fileList) {
-      await worker.addDictionarySourceToProject(file.name, file);
+    try {
+      for (let file of fileList) {
+        await worker.addDictionarySourceToProject(file.name, file);
+      }
+      error = null;
+    } catch (e) {
+      error = e;
```
ChrisChrisLoLo, line 48: I personally like to either rethrow the error or use `console.error(e)` in these cases. This specific case might be verbose enough, but I think it's a good practice to have some immediate way for the developer to get notified of the error in case the `error` variable breaks at some point

src/worker/predictive-text-studio-worker-impl.ts
```diff
@@ -82,7 +82,27 @@ export class PredictiveTextStudioWorkerImpl
     name: string,
     contents: File
   ): Promise<number> {
-    const wordlist = await readExcel(await contents.arrayBuffer());
+    let wordlist: WordList = [];
+    if (/\.(tsv)$/i.test(name)) {
```
ChrisChrisLoLo, line 86: Nitpick: we could separate the logic here from line 88 to line 99 and create a function that can be used to look like line 101: 
```ts
wordlist = await readTSV(await contents.arrayBuffer());
```
This would make addDictionarySourceToProject() more consise

src/worker/read-wordlist.ts
```diff
@@ -38,6 +38,40 @@ export async function readExcel(
   return wordlist;
 }
 
+/**
+ * Function to read a TSV file with logic from keymanapp
```
ChrisChrisLoLo, line 42: Good idea with the Doc String! These really help catch me up to speed on what the function is doing

test/test-read-wordlist.ts
```diff
@@ -24,3 +28,25 @@ test("it should remove the headers and commented rows in an Excel file", async (
     wordlistHeaderAndHashesResult
   );
 });
+
+test("it should return a wordlist given a TSV file", async (t) => {
```
ChrisChrisLoLo, line 32: Great! Thanks for adding these tests!
Refactors and new extensions should be a breeze now with these

ChrisChrisLoLo (approved):
Thanks for looking through and implementing the suggestions!

I've run the second round of review and left a few more additional suggestions.

I'll approve since I think the new suggestions are worth considering, but not worth blocking the PR. If we were working full time I'd keep this as `Request changes`, but since we're async it might be smoother to mark this as `approve`.

I'll let you use your best judgement to decide what you want to implement and what you might want to push back on. 

Feel free to leave a comment on anything if you disagree! It'd be good to discuss trade-offs if any exist

src/app/components/Upload.svelte
eddieantonio: Typically, people use a service like [Sentry](https://sentry.io/welcome/) or [Rollbar](https://rollbar.com/) to log and report this to the devs, but in my opinion,`console.error()` does not help end users!

Perhaps there's a better way to do error reporting? But that can be done in a future PR!

src/app/components/Upload.svelte
eddieantonio: I would like semantic names like `var(--error-bg-color)` and `var(--error-fg-color)` better!

src/worker/predictive-text-studio-worker-impl.ts
eddieantonio: 👍🏼  this would be a good refactor! Not required for this PR though.

src/worker/read-wordlist.ts
zprobs: word and countText are reassigned on the following lines so changing to const will cause an error. Pushing a small refactor to avoid the eslint complaints

src/worker/predictive-text-studio-worker-impl.ts
zprobs: readTSV does not take an arrayBuffer it takes a string so I had to read the string in addDictionaryToSourceProject. using FileReader. My original PR had concise notation since I just passed a Blob to readTSV but I had to change it to make the ava tests since I could not use Blobs there.

src/app/components/Upload.svelte
zprobs: Added

Issue #198, opened 2021-01-22
Test Google Sheets integration


eddieantonio:
Ensure the Google Sheets integration works with valid links and provides **helpful/actionable feedback** when things go wrong.

ChrisChrisLoLo:
@eddieantonio We might need you to create an encrypted secret so that way the Github Actions runner can use it to enable the Google Sheets API. I cannot access the settings tab for this repo, so I do not think I can tinker around with the following steps: https://docs.github.com/en/actions/reference/encrypted-secrets#creating-encrypted-secrets-for-a-repository

eddieantonio:
I added `GOOGLE_SHEETS_API_KEY` and `GOOGLE_SHEETS_CLIENT_ID` to the repository secrets 👍🏼 

You can use them in GitHub Actions as:

```yaml
env:
  API_KEY: ${ secrets.GOOGLE_SHEETS_API_KEY }
  CLIENT_ID: ${ secrets.GOOGLE_SHEETS_CLIENT_ID }
```

No need to juggle things in the `.env` file (at least, theoretically 🤞🏼 )

PR #228, opened 2021-01-24
Feat/remove headers from uploaded wordlists via the Google Sheets API

### LINKED ISSUE #223: Remove headers from uploaded wordlists via the Google Sheets API
An extension of issue #199
> Remove headers such as "word" and "count" headers in the first row of all uploaded wordlists (either Google Sheets or Excel).
> 
> This should be done automatically.

Add functionality from #217 to the Google Sheets API. Code reuse is ideal.

yehee:
Resolves #223

### DIFF
--- a/src/app/components/GoogleSheetsInput.svelte
+++ b/src/app/components/GoogleSheetsInput.svelte
@@ -45,8 +45,7 @@
         // Handle the initial sign-in state.
         updateSigninStatus(gapi.auth2.getAuthInstance().isSignedIn.get());
       })
-      .catch((err) => {
-        console.error(err);
+      .catch(() => {
         error = "Error: Could not connect to Google Sheets";
       });
   }
@@ -74,6 +73,7 @@
    *
    */
   function getSpreadsheetId() {
+    // TODO: should not use id; should make use of bindings
     googleSheetsURL = document.getElementById("input-googleSheetsURL").value;
 
     const googleSheetRegExp = RegExp(
@@ -129,8 +129,8 @@
     for (let i = 0; i < range.values.length; i++) {
       const row = range.values[i];
       if (shouldRowBeConverted(row)) {
-        const word = row[0];
-        let wordCount = row[1];
+        const word = row[WORD];
+        let wordCount = row[COUNT];
         if (!wordCount) {
           wordCount = 0;
         }
--- a/src/app/components/GoogleSheetsInput.svelte
+++ b/src/app/components/GoogleSheetsInput.svelte
@@ -6,9 +6,13 @@
   var error = null;
   let googleSheetsURL = "";
 
+  // Constants for checking headers
+  const WORD = 0;
+  const COUNT = 1;
+
   // Client ID and API key from the Developer Console
-  const CLIENT_ID = process.env.G_CLIENT_ID;
-  const API_KEY = process.env.G_API_KEY;
+  // const CLIENT_ID = process.env.G_CLIENT_ID;
+  // const API_KEY = process.env.G_API_KEY;
 
   // Array of API discovery doc URLs for APIs used by the quickstart
   const DISCOVERY_DOCS = [
@@ -45,7 +49,8 @@
         // Handle the initial sign-in state.
         updateSigninStatus(gapi.auth2.getAuthInstance().isSignedIn.get());
       })
-      .catch(() => {
+      .catch((err) => {
+        console.error(err);
         error = "Error: Could not connect to Google Sheets";
       });
   }
@@ -73,10 +78,10 @@
    *
    */
   function getSpreadsheetId() {
-    googleSheetsURL = document.getElementById("googleSheetsURL").value;
+    googleSheetsURL = document.getElementById("input-googleSheetsURL").value;
 
     const googleSheetRegExp = RegExp(
-      "http(s?)://docs.google.com/spreadsheets/[a-z]+/[a-zA-z0-9]+/[a-zA-z0-9#=]+"
+      "http(s)?://docs.google.com/spreadsheets/[a-z]{1}/[\\w-]+/[\\w#=-]+"
     );
 
     if (googleSheetRegExp.test(googleSheetsURL)) {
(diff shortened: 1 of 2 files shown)


src/app/components/GoogleSheetsInput.svelte
```diff
@@ -73,10 +73,11 @@
    *
    */
   function getSpreadsheetId() {
-    googleSheetsURL = document.getElementById("googleSheetsURL").value;
+    // TODO: should not use id; should make use of bindings
+    googleSheetsURL = document.getElementById("input-googleSheetsURL").value;
 
     const googleSheetRegExp = RegExp(
-      "http(s?)://docs.google.com/spreadsheets/[a-z]+/[a-zA-z0-9]+/[a-zA-z0-9#=]+"
+      "http(s?)://docs.google.com/spreadsheets/[a-z]{1}/[\\w-]+/[\\w#=-]+"
```
eddieantonio: Interesting! So the old regex was too restrictive? How did you come up with this regex?

Also, `{1}` is always redundant — it means select the previous token exactly once, which is the default:

```suggestion
      "http(s?)://docs.google.com/spreadsheets/[a-z]/[\\w-]+/[\\w#=-]+"
```

eddieantonio (approved):
LGTM! Just one minor suggestion, so I will let you hit the merge button when ready.

src/app/components/GoogleSheetsInput.svelte
```diff
@@ -6,9 +6,9 @@
   var error = null;
   let googleSheetsURL = "";
 
-  // Client ID and API key from the Developer Console
-  const CLIENT_ID = process.env.G_CLIENT_ID;
```
ChrisChrisLoLo, line 10: What are your thoughts on keeping this line in more or less?
I like how the naming is now `CLIENT_ID` rather than `G_CLIENT_ID`, but I liked how this line explicitly states that `CLIENT_ID` comes from an env file.

I'm neutral on this, so feel free to do what you think looks best

src/app/components/GoogleSheetsInput.svelte
```diff
@@ -6,9 +6,9 @@
...
-  const API_KEY = process.env.G_API_KEY;
+  // Constants for checking headers
+  const WORD = 0;
```
ChrisChrisLoLo, line 10: Do you think that we should make a new file for constants like these, now that it's used in multiple places?

src/app/components/GoogleSheetsInput.svelte
```diff
@@ -117,12 +128,14 @@
 
     for (let i = 0; i < range.values.length; i++) {
       const row = range.values[i];
-      const word = row[0];
-      let wordCount = row[1];
-      if (!wordCount) {
-        wordCount = 0;
+      if (shouldRowBeConverted(row)) {
```
ChrisChrisLoLo, line 131: The behaviour looks like it matches the Excel file handling, thanks!

I think we should make a new ticket to try to run all the file types into the same business rules so that we don't need to make 3+ changes for each new feature. What are your thoughts?

src/app/components/GoogleSheetsInput.svelte
ChrisChrisLoLo: Out of curiosity, what's does this different pattern do compared to the old pattern?

ChrisChrisLoLo (approved):
Looks good to me!

Great eye for detail, the code here looks very similar to the Excel reader.

I'll let you decide if you want to merge after reading my comments and suggestions. If you want a re-review feel free to let me know.

I would have requested that you added a test for this, but currently, we have not yet set that up (#198), and it could take a fair amount of time considering that we need an API key for automation. I want to avoid Unhelpful behaviour # 3 in this article: https://medium.com/@sandya.sankarram/unlearning-toxic-behaviors-in-a-code-review-culture-b7c295452a3c. We should probably get the testing going for it soon though.

src/app/components/GoogleSheetsInput.svelte
yehee: I noticed that mine includes underscores and dashes, so I've included them: `https://docs.google.com/spreadsheets/d/11NohN7t_iZ1hRx-0J7Qtr7ZMVlGl0ffjui5HLqzwhmU`. I'd love to take a look if there's documentation that I can refer back to!

src/app/components/GoogleSheetsInput.svelte
yehee: Ahh I can definitely see that! However with your documentation on client/API keys, I'm hoping that should be clear in the devs perspective, what do you think?

src/app/components/GoogleSheetsInput.svelte
yehee: Yes, but I also noticed that yours wasn't zero-based i.e. WORD = 1. So they unfortunately do not share the same value..

src/app/components/GoogleSheetsInput.svelte
yehee: It would be really nice, might need to do some investigation before since each of their implementations differs by a lot!

src/app/components/GoogleSheetsInput.svelte
yehee: This part `[a-zA-z0-9]+` from the old one was a little more strict in that it didn't allow underscores or dashes. Other changes are just to make it a bit simpler!

src/app/components/GoogleSheetsInput.svelte
ChrisChrisLoLo: Good point, I'll make some investigation ticket to start a discussion. There might be an OOP pattern that could help us

src/app/components/GoogleSheetsInput.svelte
ChrisChrisLoLo: Good catch

src/app/components/GoogleSheetsInput.svelte
ChrisChrisLoLo: Yeah, I definitely see what you're saying. Both styles can be figured out by the developer using `ctrl`+`shift`+`F`. Considered me swayed!

src/app/components/GoogleSheetsInput.svelte
eddieantonio: [Template Method Pattern!](https://en.wikipedia.org/wiki/Template_method_pattern)

Issue #54, opened 2020-09-20
Download button actually downloads the KMP file


eddieantonio:
To close this issue, we must have an end-to-end test:

**Given** all required metatadat is filled out and at least one dictionary source is uploaded
**When** I click the <kbd>Download</kbd> button
**Then** it should start a download, and I should receive a .kmp file

Helpful links:
 * https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL
 * https://dev.to/viveknayyar/e2e-testing-of-excel-downloads-with-cypress-21fb
 * https://stackoverflow.com/a/19328891/6626414

eddieantonio:
Fixed by #78

Issue #184, opened 2021-01-22
Cache Keyman API on a daily basis


eddieantonio:
Currently, we call the Keyman API to get its lists of keyboards and supported languages/local.

Our request asks `https://api.keyman.com/` for literally everything it has, which requires some heafty backend work on its part. However, the relelvant data we want changes infrequently—basically, whenever a pull request is merged into https://github.com/keymanapp/keyboards which happens a few times per week.

---

The idea is to use a [GitHub Actions scheduled job (cron job)](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#scheduled-events) to do the current request we're making on each page load. This cronjob will then save the request as JSON and **commit it to the repository** if the repsonse has changed. 

We then configure Predictive Text Studio to access the commited JSON file instead of doing the current request to the Keyman API.

**NOTE**: we may want a dedicated GitHub repository for doing the cronjob, or do it on an orphaned branch in this repository!

eddieantonio:
I marked this as `High priority` because it's really starting to annoy me now, haha!

PR #230, opened 2021-01-24
refactor: address todos

eddieantonio:
I got rid of a few `TODO` comments, as well as moved the automatic compilation/download URL logic to its own dedicated file (instead of being in `LandingPage.svelte`!)

Will make #54 easier to implement!


src/app/components/LanguageNameInput.svelte
```diff
@@ -5,33 +5,60 @@
...
+  let knownLanguages: KeyboardMetadata[] = [];
+
+  /**
+   * What text should appear below the input. In practice, this is the BCP-47
+   * tag.
+   */
+  let subtext = "";
 
-  onMount(async () => {
+  onMount(async function loadLanguageListFromWorker() {
```
zprobs, line 49: When someone first visits the site without languages cached it takes a few seconds for them to load in. Not much we can do about that except maybe add a small spinner so they know what is happening. There is a problem however when the user clicks the drop down before the languages are loaded in, they will fail to appear indefinitely.
This issue will persist on refresh so it may prevent some people from using the product, particularly those with slow internet.

zprobs (commented):
Language loading could be dealt with in a separate PR. If you'd like to merge this now let me now as the rest LGTM

src/app/components/LanguageNameInput.svelte
eddieantonio: You're right. See #184.

I'll also create a new issue about "poor UX" for when it's loading the languages.

eddieantonio: @zprobs I will merge the refactor PR, since refactors tend to breed merge conflicts, and the language loading can be improved in a new PR or two!

zprobs (approved):

PR #232, opened 2021-01-29
Remove hardcoded mockup data

### LINKED ISSUE #219: Remove hardcoded mockup data
There are a lot of hardcoded values:

- [x] ~~Languages page: language name~~ already done!
- [ ] Langugae sources: all of the data in the table
- [ ] Download button: the wordcount

These all need to be removed and replaced with the actual content in the database.

Note: it may be beneficial to resolve this issue in multiple PRs

yehee:
Remove hardcoded mockup data:
- [ ] Languages page: language name
- [x] Language sources: all of the data in the table
- [ ] Download button: the wordcount

Resolves #219

### DIFF
--- a/src/app/components/LanguageSources.svelte
+++ b/src/app/components/LanguageSources.svelte
@@ -1,13 +1,6 @@
-<script lang="ts" context="module">
-  export interface DictionarySource {
-    name: string;
-    size: number;
-    type: string;
-  }
-</script>
-
 <script lang="ts">
   import AddSource from "./AddSource.svelte";
+  import type { DictionarySource } from "../types";
   export let sources: DictionarySource[];
 
   /**
--- a/src/app/components/LanguageSources.svelte
+++ b/src/app/components/LanguageSources.svelte
@@ -1,11 +1,13 @@
-<script lang="ts">
-  import AddSource from "./AddSource.svelte";
-  interface DictionarySource {
+<script lang="ts" context="module">
+  export interface DictionarySource {
     name: string;
     size: number;
     type: string;
   }
+</script>
 
+<script lang="ts">
+  import AddSource from "./AddSource.svelte";
   export let sources: DictionarySource[];
 
   /**
--- a/src/app/pages/Languages.svelte
+++ b/src/app/pages/Languages.svelte
@@ -1,8 +1,7 @@
 <script lang="ts">
+  import type { DictionarySource } from "../types";
   import LanguageInfo from "../components/LanguageInfo.svelte";
-  import LanguageSources, {
-    DictionarySource,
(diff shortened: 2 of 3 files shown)


yehee:
@eddieantonio quick question about the DoD:
- I can't seem to find where the language name is being hardcoded, could you perhaps point me where to look?
- Seems like the `wordCount` is looking at the right `sources` which I cleared out -- it now says 0 words. I'll further test to see if it's working properly.

eddieantonio:
@yehee 

> I can't seem to find where the language name is being hardcoded, could you perhaps point me where to look?

To be honest, I don't think the language name is hardcoded anymore. I think it got fixed in a recent PR and I just didn't notice that it was already fixed. So **do not worry about that in this PR!**

> Seems like the wordCount is looking at the right sources which I cleared out -- it now says 0 words. I'll further test to see if it's working properly.

For sure! It just seems like that information is not coming from the database though :/

yehee:
@eddieantonio

> For sure! It just seems like that information is not coming from the database though :/

Right, should I create a separate issue for that? Quickly skimming through I don't think there's any GitHub issue that's addressing it.

eddieantonio:
> @eddieantonio
> 
> > For sure! It just seems like that information is not coming from the database though :/
> 
> Right, should I create a separate issue for that? Quickly skimming through I don't think there's any GitHub issue that's addressing it.

Yes! Feel free to create an issue for it.

src/app/components/LanguageSources.svelte
```diff
@@ -1,11 +1,13 @@
-<script lang="ts">
-  import AddSource from "./AddSource.svelte";
-  interface DictionarySource {
+<script lang="ts" context="module">
+  export interface DictionarySource {
```
eddieantonio: This may be out of scope for this PR, but I'd like to [keep TypeScript code in Svelte to a minimum](https://github.com/eddieantonio/predictive-text-studio/blob/production/docs/conventions.md#svelte-keep-script-sections-as-small-as-possible).

So I suggest extracting this interface to a dedicated `.ts` file, something like

```typescript
// src/app/types.ts
export interface DictionarySource { /* ... */ }
```

And then use it in Svelte files as:

```svelte
<script lang="ts">
  import type { DictionarySource } from "../types";
```

**The `import type` is important!** Otherwise, Svelte gives you a misleading error message, like "cannot import DictionarySource" or "types.ts does not export DictionarySource", something like that.

yehee: I think I've seen a few examples like this around, I'll give that a try!

eddieantonio (approved):
LGTM

PR #236, opened 2021-02-01
refactor: Upload.svelte

### LINKED ISSUE #214: Extract most of the logic in Upload.svelte
For valid code quality reasons and some silly technical limitations, we should avoid writing tonnes of TypeScript code within Svelte files.

Unfortunately, some of our components, like `Upload.svelte` have most of their logic written directly in Svelte.

Most of this code should be extracted to a dedicated `.ts` file, to facilite type-checking, correct source mapping, and code organziation in general!

eddieantonio:
I extracted a lot of the logic to `src/app/logic/upload.ts` and cleaned up the file. I also fixed a minor bug related to the handling of the `dragover` event.

Resolves #214

### DIFF
--- a/cypress/integration/landingPage.spec.ts
+++ b/cypress/integration/landingPage.spec.ts
@@ -40,7 +40,6 @@ describe("Upload from the the landing page", () => {
   it("should disable to continue if only file is provided", () => {
     // we need a valid bcp47Tag to update the packageInfo
     cy.data("autocomplete-label").scrollIntoView().should("be.visible");
-    cy.data("autocomplete-label").type("zh");
 
     const downloadedFilePath = path.join(downloadFolder, "Example.kmp");
 
--- a/cypress/integration/landingPage.spec.ts
+++ b/cypress/integration/landingPage.spec.ts
@@ -38,9 +38,6 @@ describe("Upload from the the landing page", () => {
   });
 
   it("should disable to continue if only file is provided", () => {
-    // we need a valid bcp47Tag to update the packageInfo
-    cy.data("autocomplete-label").scrollIntoView().should("be.visible");
-
     const downloadedFilePath = path.join(downloadFolder, "Example.kmp");
 
     cy.data("quick-start")
--- a/src/app/logic/upload.ts
+++ b/src/app/logic/upload.ts
@@ -16,6 +16,19 @@ export function getAllFilesFromDragEvent(event: DragEvent): File[] {
 }
 
 /**
+ * Returns an array of files from the given event target, which is assumed to
+ * be an <input type="file">
+ */
+export function filesFromInputElement(el: EventTarget | null): File[] {
+  if (el === null) return [];
+  const input = el as HTMLInputElement;
+
+  if (!input.files) return [];
+
+  return Array.from(input.files);
+}
+
+/**
  * Uploads all the given files to the current project.
  */
 export async function addAllFilesToCurrentProject(
--- a/src/app/logic/upload.ts
+++ b/src/app/logic/upload.ts
@@ -1,0 +1,17 @@
+/**
+ * Return File objects from a data transfer list (created by some drag
+ * events).
+ */
+export function fileFromDataTransferItem(items: DataTransferItemList): File[] {
+  const fileList: File[] = [];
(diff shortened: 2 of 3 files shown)


src/app/logic/upload.ts
```diff
@@ -0,0 +1,74 @@
...
+  }
+}
+
+/**
+ * Returns an array of files from the given event target, which is assumed to
+ * be an <input type="file">.
+ */
+export function filesFromInputElement(el: EventTarget | null): File[] {
+  if (el === null) return [];
+  if (!isFileInputElement(el)) return [];
```
ChrisChrisLoLo, line 24: What are your thoughts on adding some sort of warning or error if the input is null or not the file input element?

cypress/integration/landingPage.spec.ts
```diff
@@ -38,10 +38,6 @@ describe("Upload from the the landing page", () => {
   });
 
   it("should disable to continue if only file is provided", () => {
-    // we need a valid bcp47Tag to update the packageInfo
-    cy.data("autocomplete-label").scrollIntoView().should("be.visible");
```
ChrisChrisLoLo, line 42: Do these lines cause problems with the newly refactored code?

ChrisChrisLoLo (approved):
LGTM :shipit:!

Added two comments. Feel free to merge after reading them

cypress/integration/landingPage.spec.ts
eddieantonio: Yes. Although, the thing about refactoring is that, if you've done it right, it should cause no effect on the tests.

I'm honestly pretty confused on why this was working before. The issue here is that typing "zh" in the language input would open the autocomplete selection box, which would then obscure the drag-and-drop area, making the rest of the test fail because the file upload area is inaccessible. However, this test "it should disable to continue if only file is provided" should not even _touch_ the language input... so why... was it doing this stuff in the first place?

src/app/logic/upload.ts
eddieantonio: I have complicated thoughts on this!

In general, my answer is yes, errors are good. In particular, this _needs_ to be called on a `<input type="file">` element, so both `if (el === null)` and `if (!FileInputElement())` should raise some sort of assertion error (a.k.a., programmer's assumptions are invalid). However, who would catch that error? This code is destined for the app bundle, so logging via `console.error()` would only serve to clutter up users' console logs, that most users won't see. If we were using some crash reporting (e.g., Sentry or Rollbar), we would send that error here.

Haha, so with that 4 dimensional chess out of the way, I decided just to forget about raising error messages, because errors in browser code are annoying :/

ChrisChrisLoLo: That explanation makes sense to me and is useful for gaining insight into front-end development, thanks :+1:!

If you don't mind, I'm curious about my current assumptions about using the browser console. My current rationale in favour of using the console is exactly because most users can't see it, so that developers can take a peek when intermediate processes have gone awry to save time debugging. I agree that it's never good when the console is filled with needless logs, but do you think using `console.error()` sparingly is acceptable? I'm curious about this subject as it's one I don't have a lot of experience in.

As this project expands in complexity in its logic, do you think we should add a ticket to add something like Sentry or Rollbar? If not crash reporting or logging, is there some other solution for flagging intermediate errors to help developers pinpoint bugs?

Once again, I appreciate picking your brain on these subjects!

eddieantonio: > I agree that it's never good when the console is filled with needless logs, but do you think using console.error() sparingly is acceptable? I

I don't have a clear answer for this, unfortunately. My biggest issue against it is that [`console.error()` is a memory leak](https://cherniavskii.com/be-careful-with-console-logs/#console-warnings). But yes, using it sparingly — if you can guarentee it once per page load — is acceptable.

> As this project expands in complexity in its logic, do you think we should add a ticket to add something like Sentry or Rollbar? If not crash reporting or logging, is there some other solution for flagging intermediate errors to help developers pinpoint bugs?

Sure! I'm not sure about other solutions for flagging intermediate errrors. There are static analysis tools, including the TypeScript typechecker itself. But you have to write clever code in TypeScript to be able to take advantage of it.

PR #240, opened 2021-02-02
fix: upload integration test

eddieantonio:
The `upload.spec.ts` test case was disabled due to the download button being removed from the landing page. Although the landing page should be simplified, the upload test is super important, so I figured it's worth making the landing page a bit sloppy while we figure out where the download button should actually go!


zprobs (approved):
Downloading works as expected. The download button is still enabled if the user does not enter a language. Not sure if this is desired behaviour or not but will leave that up to you.

Issue #235, opened 2021-02-01
Create "source type" union in @common/types; store in DB


eddieantonio:
We have a few source types:

 * .xlsx
 * Google Sheets
 * TSV
 * Direct entry

We should have a type, something like the following:

```typescript
type DictionarySourceType = "xlsx" | "google-sheets" | "tsv" | "direct-entry";
```

And we should store this in the database whenever we store a dictionary source. This can then be converted to a nice, human readable string in the "dictionary sources" panel.

PR #246, opened 2021-02-09
Feat/load sources from db

### LINKED ISSUE #234: Information on Languages page should be loaded from the database
- Per discussion on #232, referring to [this comment](https://github.com/eddieantonio/predictive-text-studio/blob/e91b922935c251df5a51ee68fa802580a614521b/src/app/pages/Languages.svelte#L9)

yehee:
### Changes
- Define `getFilesFromStorage` to fetch all files from the storage
- Change `saveFile(name, wordlist)` to `saveFile(StoredWordList)` to include metadata like file size, file type
- Add `size` and `type` attributes to `StoredWordList`
- Update test cases to reflect changes

### Discussion
- **Should we flush out the test files?** Right now any source that are missing size/type have `undefined` values
- Should we have an indicator when there are no sources/files to list?

Fixes #234 

Screen Shot 2021-02-09 at 11 28 54 AM


eddieantonio:
> * **Should we flush out the test files?** Right now any source that are missing size/type have `undefined` values

We're early enough in development that flushing test values is probably for the best! I never want to see `undefined` in a user interface!

> * Should we have an indicator when there are no sources/files to list?

Yes! Do you think that's a quick thing to add to this PR, or should we make a seperate issue for that?

src/app/types.ts
```diff
@@ -1,5 +1,27 @@
...
+  wordlist: WordList;
+  /**
+   * Size of the file in bytes.
+   */
+  size?: number;
+  /**
+   * File extension.
+   */
+  type?: string;
+}
```
eddieantonio: This looks like a copy-paste from`@worker/storage-types`? It's better to maintain a single copy of this class, and that copy should be in `@common/types` — so please move the interface to `@common/types`!

Is the `id?` required information for the UI? Or is it exclusive to the storage layer (Dexie/IndexedDB)? In which case, it should not be in the `@common/types` or even `@app/types` file!

You can do something like this. In `@common/types`:

```typescript
export interface WordlistSource {
  /* ... all properties EXCEPT id */
}
```

Then in `@worker/storage-types`:

```typescript
export interface StoredWordlist extends WordlistSource {
  /** Autogenerated primary key */
  id?: number;
}
```

src/worker/storage-models.ts
```diff
@@ -19,6 +19,14 @@ export interface StoredWordList {
...
    */
   wordlist: WordList;
+  /**
+   * Size of the file in bytes.
+   */
+  size?: number;
+  /**
+   * File extension.
+   */
+  type?: string;
```
eddieantonio: Please make them MANDATORY!

```suggestion
  /**
   * Size of the file in bytes.
   */
  size: number;
  /**
   * File extension.
   */
  type: string;
```

You may need to write an [upgrader function](https://dexie.org/docs/Version/Version.upgrade()#sample), that flushes the database as you mentioned.

src/worker/storage.ts
```diff
@@ -114,10 +113,10 @@ export default class Storage {
   /**
    * Saves a wordlist source file to storage.
    */
-  saveFile(name: string, wordlist: WordList): Promise<void> {
+  saveFile(file: StoredWordList): Promise<void> {
```
eddieantonio, line 122: I like this change in the API! 👍🏼

src/app/types.ts
```diff
@@ -1,5 +1,27 @@
...
+  /**
+   * Typically the filename of the uploaded dictionary source.
+   */
+  name: string;
+  /**
+   * The actual contents of said file.
+   */
+  wordlist: WordList;
+  /**
+   * Size of the file in bytes.
```
eddieantonio: **words** is a much more useful thing to store than bytes, for the end-user perspective!

```suggestion
   * Size of the file in words.
```

src/worker/storage-models.ts
eddieantonio: Please refer to #235.

To avoid scope-creep in this PR, feel free to **omit `type` entirely** and leave it for a future PR! 😅

test/test-storage.ts
```diff
@@ -70,16 +73,16 @@ test("retrieving mulitple files with .fetchAllFiles()", async (t) => {
 
   sources.sort(byName);
 
-  for (const { name, wordlist } of sources) {
-    await storage.saveFile(name, wordlist);
+  for (const source of sources) {
```
eddieantonio, line 90: This is so much nicer!

src/worker/storage-models.ts
```diff
@@ -2,7 +2,7 @@
  * @file Interfaces used in storage
  */
 
-import { WordList, ProjectMetadata } from "@common/types";
+import type { WordList, ProjectMetadata } from "@common/types";
```
eddieantonio: FYI: `import type` is _not_ required in regular `.ts` files. It's just a hack that's required for `.svelte` files :/

However... it is nicely self-documenting... so maybe keep this line as it is! 😄

eddieantonio (changes requested):
This is good stuff!

I have concerns about duplicate interface definitions, and I'm realziing that incorporating `size`  (in words) and `type` could be separate issues entirely. You can just write "unknown" in the interface for now, and we can handle adding `size` and `type` to future PRs to avoid ballooning this PR!

See also: #235.

src/worker/storage-models.ts
yehee: @eddieantonio #235 seems to be addressing the exact issue discussed here! I'll pick up on that issue once I wrap up this PR.

yehee: I still have `type` as optional, but I'll be working on it right away to address it. Would that be ok with you?

yehee: I've manually cleaned up files cached in the DB for now, where would the upgrader function go if they would be implemented and when should we call it?

src/app/types.ts
yehee: I think this would work the best! I'll update the PR accordingly. Thanks!

src/worker/storage-models.ts
eddieantonio: If you're going to work on #235 directly after this, then, yes, that's fine!

---

I just noticed that this PR doesn't actually update the database schema, but if it's going to change the `StoredWordlist` interface, then it must!

**EDIT**: I have an idea about how to fix this _without_ schema migrations. Will post in another comment.

---

Here is where the schema declaration and update function would go: https://github.com/eddieantonio/predictive-text-studio/blob/764510b168f2e6eacc1ad3b3552687162b680995/src/worker/storage.ts#L91

I just realized that I misconstrued the `SCHEMA_VERSION` constant. It turns out that if you need to change how data is represented, the Dexie schema must have `this.version(...)` lines. So here's what I'm proposing:

Delete the `SCHEMA_VERSION` constant, then add each schema revision:

```typescript
    this.version(4).stores({
      /* "Initial" schema: */
      files: "++id, name, wordlist",
      projectData:   "++id, langName, bcp47Tag, authorName, modelID, copyright, version",
      keyboardData: "++id, bcp47Tag, language, timestamp",
      KMPFileData: "++id, package",
    });

    /* Version 5: Add "size"  property to file table: */
    this.version(5).stores({
      files: "++id, name, wordlist, size"
    }).upgrade((transaction) => {
      return transaction.table("files").toCollection().modify((file: StoredWordlist) => {
        file.size = file.wordlist.length
      });
    });
```

Please see this documentation: https://dexie.org/docs/Tutorial/Design#database-versioning

eddieantonio (changes requested):
It's looking good!

However, the additional properties in `StoredWordlist` (namely, `size` and `type`) are a problem. Since old versions of the database on other peoples' computers do not have these properties, this will cause all kinds of inconsistencies in the rest of the code.

# Solution 1

One thing we can do is... ignore it. We don't have many users, and if things break, we can tell them to clear their browser's IndexedDB. This is okay now, but it's not a habit you want to make as a professional software developer! 🙈 

# Solution 2

Add a database migration (see [this comment](https://github.com/eddieantonio/predictive-text-studio/pull/246#discussion_r575337212)).

# Solution 3

Since only `size` is mandatory and it's a _dervied property_, then we can just add it:

```typescript
function enrichWordlist(source: Partial<WordListSource>): WordListSource {
   if (!source.name || !source.wordlist) throw new Error("missing mandatory properties!")
   return {
     name,
     wordlist,
     size: wordlist.length,
     // TODO: could even add type here by inferring it from the `name`
   };
}
```

(See documentation on [`Partial<Type>`](https://www.typescriptlang.org/docs/handbook/utility-types.html#partialtype))

Feel free to DM me on Slack if any of this is unclear!

yehee:
@eddieantonio Thanks a lot for putting these altogether! I went with the 2nd approach as it seems like with this approach adding `type` would also be quite straightforward. Let me know if you see anything misleading from the changes.

eddieantonio (approved):
LGTM! Thanks for working on this!

PR #247, opened 2021-02-13
Feat/create "source type" union in @common/types; store in DB

### LINKED ISSUE #235: Create "source type" union in @common/types; store in DB

yehee:
- Define `type DictionarySourceType` in `@common/types`:
```js
type DictionarySourceType = "xlsx" | "google-sheets" | "tsv" | "direct-entry";
```
- Add `type` as `DictionarySourceType` in `WordListSource`

Resolves #235 

**Note:**
- Tested with `.xlsx` and `tsv`. Direct entry and Google Sheets need to be tested further once they are functional.

Screen Shot 2021-02-12 at 11 00 35 PM

### DIFF
--- a/src/app/components/LanguageSources.svelte
+++ b/src/app/components/LanguageSources.svelte
@@ -19,7 +19,7 @@
    */
   const handleDelete = (): void => {};
 
-  const dict = {
+  const englishNameOf = {
     "direct-entry": "Direct entry",
     "google-sheets": "Google Sheets",
     tsv: "TSV",
@@ -108,7 +108,7 @@
       <tr>
         <td>{source.name}</td>
         <td>{source.size}</td>
-        <td>{dict[source.type]}</td>
+        <td>{englishNameOf[source.type]}</td>
         <td class="table__row--actions actions">
           <button
             class="actions__action btn--inline"
--- a/src/app/components/LanguageSources.svelte
+++ b/src/app/components/LanguageSources.svelte
@@ -18,6 +18,13 @@
    * @return {void}
    */
   const handleDelete = (): void => {};
+
+  const dict = {
+    "direct-entry": "Direct entry",
+    "google-sheets": "Google Sheets",
+    tsv: "TSV",
+    xlsx: ".xlsx",
+  };
 </script>
 
 <style>
@@ -101,7 +108,7 @@
       <tr>
         <td>{source.name}</td>
         <td>{source.size}</td>
-        <td>{source.type}</td>
+        <td>{dict[source.type]}</td>
         <td class="table__row--actions actions">
           <button
             class="actions__action btn--inline"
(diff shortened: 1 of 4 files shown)


src/worker/predictive-text-studio-worker-impl.ts
```diff
@@ -105,12 +111,19 @@ export class PredictiveTextStudioWorkerImpl
         reader.readAsText(contents);
       });
       wordlist = readTSV(TSVFileString);
+      type = "tsv" as DictionarySourceType;
```
```suggestion
      type = "tsv";
```

src/worker/predictive-text-studio-worker-impl.ts
```diff
@@ -105,12 +111,19 @@ export class PredictiveTextStudioWorkerImpl
...
     } else if (/\.(xlsx)$/i.test(name)) {
       wordlist = await readExcel(await contents.arrayBuffer());
+      type = "xlsx" as DictionarySourceType;
```
```suggestion
      type = "xlsx";
```

src/app/components/LanguageSources.svelte
```diff
@@ -18,6 +18,13 @@
    * @return {void}
    */
   const handleDelete = (): void => {};
+
+  const dict = {
```
eddieantonio: This constant could have a more descriptive name. How about `englishNameOf`?

```suggestion
  const englishNameOf = {
```

Feel free to come up with a different name for this, but `dict` is too generic!

src/app/components/LanguageSources.svelte
```diff
@@ -101,7 +108,7 @@
       <tr>
         <td>{source.name}</td>
         <td>{source.size}</td>
-        <td>{source.type}</td>
+        <td>{dict[source.type]}</td>
```
eddieantonio: (see above)

```suggestion
        <td>{englishNameOf[source.type]}</td>
```

test/test-storage.ts
```diff
@@ -44,6 +44,7 @@ test("storing a file", async (t) => {
     name: "ExampleWordlist.xlsx",
     wordlist: exampleWordlist,
     size: exampleWordlist.length,
+    type: "xlsx" as DictionarySourceType,
```
```suggestion
    type: "xlsx",
```

test/test-storage.ts
```diff
@@ -58,6 +59,7 @@ test("retrieving one file with .fetchAllFiles()", async (t) => {
     name: filename,
     wordlist: exampleWordlist,
     size: exampleWordlist.length,
+    type: "xlsx" as DictionarySourceType,
```
```suggestion
    type: "xlsx",
```

eddieantonio (changes requested):
Looking good! Just a few code maintainability changes, then LGTM!

Note that you don't always need to cast a string to `as DictionarySourceType`. Generally, you should avoid casting as it can mask errors, like typos!

**EDIT**: @yehee, please `git pull` this branch again if you work locally as I made a slight modification to `test-storage.ts`

eddieantonio (approved):
LGTM! :shipit:

PR #250, opened 2021-02-15
fix: generate KMP after google sheets imported

### LINKED ISSUE #198: Test Google Sheets integration

yehee:
- Tested and verified that the entries from Google Sheets URL are correctly imported and parsed
- Generate KMP file after entries are imported from Google Sheets URL so that `Customise` button is enabled

Fixes #198 

Screen Shot 2021-02-14 at 10 27 38 PM
Screen Shot 2021-02-14 at 10 28 17 PM


eddieantonio (changes requested):
I like the changes, but it does not resolve #198, as that really requires some sort of automated testing.

As for _how_ to do that automated testing... oof, I'm not really sure of an easy way of doing that :/

I'm good with merging this in, as long as #198 remains open.

eddieantonio:
@yehee once #254 is merged, then this PR can gain an intergration test that uses the Google Sheets integration and check if the `.kmp` file is valid.

yehee:
@eddieantonio I see! So once #254 is merged, I'll test to make sure the generated KMP file is sound. Would that meet the DoD for #198?

eddieantonio:
> @eddieantonio I see! So once #254 is merged, I'll test to make sure the generated KMP file is sound. Would that meet the DoD for #198?

Yes! "Test" means "automated test", unless otherwise specified. While we're clarifying terms, when you say "DoD" you mean "Definition of Done"?

yehee:
@eddieantonio Yes! In the meantime, I've merged `test/read-kmp` into this branch to test locally, and seems like it's working fine.

eddieantonio:
@yehee I've merged #254, so this is no longer blocked!

yehee:
@eddieantonio Cool! Would this be good to be merged then?

eddieantonio:
> @eddieantonio Cool! Would this be good to be merged then?

No, haha! 😄The original issue #198 is titled for "Test Google Sheet integration" which I intended to mean, "add automated test that makes a call to the Google Sheets API". In other words, if this PR wishes to close #198, it would have to include an integration test (with Cypress) that:

 1. checks to see that no .kmp file has been downloaded
 2. uses the Google Sheets API to upload a spreadsheet
 3. clicks download
 4. uses `cy.readZip()` to check that a KMP file was made (you can pretty much copy paste the assertion from `upload.spec.js` for now)

yehee:
@eddieantonio Ohh I see! I think I have the cypress test somewhat compiled together, I'm currently looking into how to handle Google authentication (or workaround with cookies?). `googlesheets.spec.ts` is failing due to this error. Please let me know if you have any suggestions wrt this!

eddieantonio:
Hey @yehee! I'm investigating the issue and there are a few problems:

 1. we're not exporting the `CLIENT_ID` and `API_KEY` in the GitHub actions tests 🙈 
 2. The client keys I've made are not setup to allow requests from "localhost:5000"

**EDIT**: here's the exact error I'm getting locally:

```
Not a valid origin for the client: http://localhost:5000 has not been whitelisted for client ID <REDACTED>. Please go to https://console.developers.google.com/ and whitelist this origin for your project's client ID.
```

These are the issues I've identified thus far. I'm going to try to resolve this. Is it okay if I push directly to this branch?

yehee:
@eddieantonio Please do! I was looking into [cypress-social-logins](https://www.npmjs.com/package/cypress-social-logins) package to see if it would help, but is there anything else that I should be looking at at the moment?

eddieantonio:
> @eddieantonio Please do! I was looking into [cypress-social-logins](https://www.npmjs.com/package/cypress-social-logins) package to see if it would help, but is there anything else that I should be looking at at the moment?

Oh, good catch! I don't think we need `cypress-social-logins` right now, but might in the future. I'll keep it in mind!

eddieantonio:
> @eddieantonio Please do! I was looking into [cypress-social-logins](https://www.npmjs.com/package/cypress-social-logins) package to see if it would help, but is there anything else that I should be looking at at the moment?

Upon a closer look, it seems like we will absolutely need [cypress-social-logins](https://www.npmjs.com/package/cypress-social-logins) to test this 😅 

I've made a few changes to the Google API configuration and made sure the keys are visible to Cypress in continuous integration. I full expec the latest commit to fail, however.

@yehee: please try to integrate [cypress-social-logins](https://www.npmjs.com/package/cypress-social-logins) right now, but I will continue investigating later this week. Let me know if you need anything else from me!

eddieantonio:
😩 

@yehee: okay, so I read literally the first paragraph of that Cypress plugin:

> This plugin doesn't work well in a CI environment, due to the anti-fraud detection mechanisms employed by the likes of Google, GitHub etc. Why? If you attempt to login from a CI machine which has different IPs, geolocation and other fingerprint identification which the account you use isn't normally attempting a login from, then this will trigger Multi Factor Authentication, CAPTCHA, or other means of confirming the identity. When those extra steps are needed, this plugin doesn't work well around them.

So... this is a dead-end :/

I think it's okay to mark this test as "works only locally" and skip it in CI. I can give guidance on how to do that later, but in order to make sure this PR doesn't balloon out of control, I'd suggest running this only on localhost (see: https://github.com/cypress-io/cypress-skip-test#cyonlyon), since... oof, it may never work in CI 🙃

yehee:
@eddieantonio Thanks for looking into this further! I definitely see how that would raise some concerns on security... I will set up the test so that it only runs locally and hopefully have a protocol set up so that whoever touches on Google Sheets API integration in their future work has some test suites that they can refer back to.

Edit: @eddieantonio Seems like this may also be challenging as well since we are working with a popup window that is not yet supported in cypress ([related discussion on StackOverflow](https://stackoverflow.com/questions/47805037/angular-cypress-io-test-with-google-login-popup-window) and [GitHub](https://github.com/cypress-io/cypress/issues/4468)). Wondering if it would at all be feasible to set up an automated test for this feature, perhaps bypassing the authentication part?

package.json
```diff
@@ -16,6 +16,7 @@
     "prettier": "prettier --write '{src,test,cypress}/**/*.{ts,svelte}'",
     "start": "sirv public -s",
     "test": "ava",
+    "test:ci": "ava --match='!*Google*'",
```
yehee, line 19: To exclude Google sheets testing on CI/production.

yehee:
@eddieantonio I've done some investigations on a couple of options I thought were viable.

1) `cypress` testing turned out to be challenging since it does not allow us to do the authentication via a third-party auth service.
2) getting access token via terminal interactively while running AVA tests locally is not an option since it doesn't seem to allow us from entering input while the test cases are being executed. I tried using `readline` to prompt the user to type in, but it doesn't seem to work.
3) This is the approach I ended up with, testing via AVA. This however requires some set-ups prior to running the test, mainly to bypass the authentication via UI. To help with the setup, I've updated the end-to-end test section of the `CONTRIBUTING.md` doc to add a few comments about how/where to get the access tokens. The instruction is still work-in-progress, I'd appreciate it if anyone has any suggestions/comments!

I've done quite a bit of refactoring, please let me know if this at all works for the application!

.gitignore
```diff
@@ -36,8 +36,8 @@ cypress/downloads/
 .DS_Store
 
 # Google Sheets Credentials
-credentials.json
-token.json
+**/credentials.json
+**/token.json
```
eddieantonio:
```suggestion
test/fixtures/credentials.json
test/fixtures/token.json
```

src/app/components/GoogleSheetsInput.svelte
```diff
@@ -54,23 +49,27 @@
    *  Called when the signed in status changes, to update the UI
    *  appropriately. After a sign-in, the API is called.
    */
-  function updateSigninStatus(isSignedIn) {
+  async function readGoogleSheet(isSignedIn) {
```
eddieantonio, line 52: This is a better name!

src/worker/read-wordlist.ts
```diff
@@ -103,3 +103,37 @@ export function readManualEntryData(contents: DictionaryEntry[]): WordList {
   });
   return wordlist;
 }
+
+/**
+ * Returns a boolean if the row should be converted.
+ * The row should not be converted if it is a header row or a commented out row.
+ */
+export function shouldRowBeConvertedGoogleSheets(row: string[]): boolean {
```
eddieantonio: Does this function _need_ to be exported?

test/test-google-sheets-api.ts
```diff
@@ -0,0 +1,87 @@
...
+// The file token.json stores the user's access and refresh tokens.
+const TOKEN_PATH = pathToFixture("token.json");
+
+/**
+ * Create an OAuth2 client with the given credentials, and then execute the
+ * given callback function.
+ * @param {Object} credentials The authorization client credentials.
+ * @param {function} callback The callback to call with the authorized client.
+ */
+function authorize(credentials: any, callback: any) {
```
eddieantonio: Please use proper types in this file rather than `any`. You do not need to specify the entire interface; just the parts the interfaces that are used.

src/worker/predictive-text-studio-worker-impl.ts
```diff
@@ -64,17 +69,10 @@ export class PredictiveTextStudioWorkerImpl
...
-    wordlist: WordList
-  ): Promise<ArrayBuffer> {
-    this.storage.saveFile({
-      name,
-      wordlist,
-      size: wordlist.length,
-      type: "google-sheets",
-    });
-    return await linkStorageToKmp(this.storage);
+  async readGoogleSheet(name: string, rows: string[][]): Promise<void> {
```
eddieantonio, line 66: Good refactor!

test/test-google-sheets-api.ts
```diff
@@ -0,0 +1,87 @@
+import test from "ava";
+import * as fs from "fs";
+import * as path from "path";
+import * as dot from "dotenv";
+import * as c from "chalk";
+import { google } from "googleapis";
```
eddieantonio, line 6: How does this get imported? Am I not seeing something in the `package.json`?

eddieantonio (changes requested):
Please add typing to the tests. Otherwise, good compromise! 😅

test/test-google-sheets-api.ts
yehee: It was in the `package.json` so we don't see it added in this PR! https://github.com/eddieantonio/predictive-text-studio/blob/production/package.json#L68 ([related PR](https://github.com/eddieantonio/predictive-text-studio/commit/d8362095e40c71c08b25af13a3e4978bd3ff5d10#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519))

test/test-google-sheets-api.ts
yehee: For sure! I think I could come up with an interface for `credentials`, but I was having trouble with the `callback` param. Would you have any suggestions?

test/test-google-sheets-api.ts
eddieantonio: oh derp, thanks for pointing that out!

test/test-google-sheets-api.ts
eddieantonio: I was looking at the `googleapis` library, and even though [it's all TypeScript](https://github.com/googleapis/google-api-nodejs-client), it's a mess! I'll see if I can come up with something.

.github/CONTRIBUTING.md
```diff
@@ -108,6 +108,44 @@ With the server running as above (with `yarn start`), run this to open
...
+
+```json
+{
+    "client_secret": YOUR_CLIENT_SECRET,
+    "client_id": YOUR_CLIENT_ID,
+    "redirect_uri": "http://localhost:5000"
+}
+```
+
+`token.json`: run `yarn test` to get the authentication URL e.g. ``. You need to authenticate yourself to get the code. Once you have the code,
```
eddieantonio, line 129: @yehee Maybe I'm missing something, but where is `credentials.json` actually _used_?

test/test-google-sheets-api.ts
```diff
@@ -0,0 +1,87 @@
...
+            t.not(rows.length, 0);
+            const { name, wordlist, size, type } = await readGoogleSheet(
+              spreadsheetId,
+              rows
+            );
+            t.is(name, spreadsheetId);
+            t.deepEqual(wordlist, [
+              ["hello", 1],
+              ["world", 2],
+            ]);
```
eddieantonio: These assertions assume a **specific** spreadsheet. What spreadsheet is it? Could you document it?

Also, for an application that is intended for ANY language, please try to avoid English words if you can! Use Korean or French. I will use Indigenous languages of Turtle Island or Spanish.

eddieantonio (approved):
LGTM!

We can improve this in the future, but it works for now. @yehee, please check out the changes on your own machine and run the tests, making sure it all passes. Once it does, feel free to hit that merge button! :shipit:

Issue #251, opened 2021-02-16
Make Language and Dictionary Name Persistant


zprobs:
Currently, when user enters a language on the landing page it will not appear on the languages page.

When switching between sources and information, the language and dictionary name data is erased.

PR #252, opened 2021-02-16
Project Compilation Timing Checks

### LINKED ISSUE #164: Initiate the package compilation at the correct time
**EDIT**: I wrote this issue poorly the first time, so I've rewritten some parts:

The package should be recompiled when:

 * a file is uploaded
 * when a metadata field (name, language, copyright, etc.) is _finished_ being modified (the "blur" event on the input box)
 * when the manual entry dialog is closed

Upon [package compilation start](https://github.com/eddieantonio/predictive-text-studio/blob/production/src/common/predictive-text-studio-worker.ts#L43-L58), the download button should be **greyed out**. Upon [package compilation success](https://github.com/eddieantonio/predictive-text-studio/blob/ccb8d62d1e6392fd011bba13098004d2de5452d2/src/common/predictive-text-studio-worker.ts#L54-L58), the package download icon should return to normal. The word count should also be displayed and the language should be updated!

zprobs:
Adding checks to the download button so it is only enabled when the KMP file is compiled and ready for download

Fixes #164


eddieantonio (changes requested):
I don't believe this fixes #164.

That issue states that the package should be recompiled under a number of different conditions. I wrote:

> ready to download when...

Which is misleading — sorry!

---

What I intended with that issue is that the download button should respond to the callbacks here:

https://github.com/eddieantonio/predictive-text-studio/blob/production/src/common/predictive-text-studio-worker.ts#L43-L58

That is, the button should be greyed out `onPackageCompileStart` and only enabled `onPackageCompileSuccess`

---

I'll update #164 to make it more clear!

zprobs:
Ok making those changes now

zprobs:
Fixes #251

src/worker/predictive-text-studio-worker-impl.ts
```diff
@@ -77,7 +77,7 @@ export class PredictiveTextStudioWorkerImpl
     return await linkStorageToKmp(this.storage);
   }
 
-  private async generateKMPFromStorage(): Promise<void> {
```
eddieantonio: Hmm :/ I'd rather not make this public and add to the footprint of the `PredcitveTextStudioWorker` public interface.

Since the app code calls this directly after `setProjectData()` (which is already part of the interface), I think it would be better for `PredictiveTextStudioWorkjerImpl.setProjectData()` to generate the KMP.

eddieantonio (changes requested):
Please see my inline comments.

I would also like some tests to check for the correct behaviour.

zprobs:
> Please see my inline comments.
> 
> I would also like some tests to check for the correct behaviour.

I started a test for compile checking in test-worker.ts but I am unable to get the test worker to compile a KMP. Please let me know what I should do to get the test worker to compile a file.

src/app/logic/automatic-compilation.ts
```diff
@@ -14,6 +14,22 @@ export function setupAutomaticCompilationAndDownloadURL(): void {
...
+  worker.onPackageCompileStart(
+    Comlink.proxy(() => {
+      compileSuccess.set(false);
+    })
+  );
+  worker.onPackageCompileSuccess(
+    Comlink.proxy(() => {
+      compileSuccess.set(true);
+    })
+  );
```
eddieantonio: So I don't believe the worker allows for multiple `onPackageCompileStart` callbacks, which means this _overwrites_ the existing `onPackageCompileSuccess` callback; namely, the one that sets up the download button.

The solution is to merge the logic here into `setupAutomaticCompilationAndDownloadURL()` and remove the `setupCompilationSuccessSignal()` function. There's no harm in setting the stores even if there are no listeners to the store's changes.

src/app/components/LanguageInfo.svelte
```diff
@@ -75,22 +95,23 @@
...
-      on:message={updateMetadata} />
+      bind:selectedLanguage={languageInfo} />
     <InputField
       label="Author or Organization"
       subtext="Shortcode"
       id="authorName"
       cyData="input-author-name"
       bind:inputValue={authorName}
-      on:message={updateMetadata} />
+      on:message={onBlurListener} />
```
eddieantonio, line 105: 👍🏼

src/app/pages/Languages.svelte
```diff
@@ -23,8 +28,21 @@
...
     languageInformation.sources = await worker.getFilesFromStorage();
+  }
+
+  // listen to changes to the package compilation and recalculate word count
+  compileSuccess.subscribe((currentValue) => {
+    downloadReady = currentValue;
+    if (currentValue) {
+      getLanguageSources();
+    }
+  });
```
eddieantonio: As mentioned in these docs <https://svelte.dev/tutorial/auto-subscriptions>, there is a sublte bug here; `.subscribe()` returns an `unsubscribe()` function that must be called in the `onDestroy()` hook; however, if you use the `$compileSuccess` syntax, Svelte will automatically subscribe and unsubscribe for you.

```suggestion
  $: downloadReady = $compileSuccess;
```

src/app/pages/Languages.svelte
```diff
@@ -23,8 +28,21 @@
...
-  onMount(async () => {
+  async function getLanguageSources() {
     languageInformation.sources = await worker.getFilesFromStorage();
+  }
+
+  // listen to changes to the package compilation and recalculate word count
+  compileSuccess.subscribe((currentValue) => {
+    downloadReady = currentValue;
+    if (currentValue) {
+      getLanguageSources();
```
eddieantonio: Why does `getLanguageSources()` get called here? As far as I understand, it should only be called `onMount()` to the list of possible languages ONCE.

eddieantonio (changes requested):
Since `setupCompilationSuccessSignal()` overwrites the existing `setupCompilationSuccessSignal()`, the code as it is now will induce a bug. I believe the easiest fix is to setup both `currentDownloadURL` _and_ `compileSuccess` in `setupAutomaticCompilationAndDownloadURL()`.

Sorry for making this PR go for so long 😭

src/app/pages/Languages.svelte
zprobs: This was to ensure that the word count was up to date when re-compilation occurred but perhaps it is better to only update the word count when a dictionary is added. I can pass `getLanguageSources()` down to the source components instead.

src/app/logic/automatic-compilation.ts
zprobs: Good catch

src/app/pages/Languages.svelte
eddieantonio: Ah, okay! There's a better way to ensure this works, but that's definitely work for a future PR 😅

eddieantonio (approved):
LGTM!

PR #258, opened 2021-02-25
feat: Add support for mobile devices to landing page

### LINKED ISSUE #241: Add support for mobile devices to landing page
Landing page is unusable on phones.

yehee:
## Changes
- Minor UI enhancement for a mobile-friendly experience

## Note
- I have introduced a couple of constants e.g. `max-width: 1024px` as in laptop screen size, `max-width: 768px` as in tablet screen size. If there are better ways to address those values, I'd like to know!

Resolves #241


src/app/components/Button.svelte
```diff
@@ -22,6 +22,13 @@
     flex-direction: column;
     align-items: center;
   }
+
+  @media (max-width: 425px) {
```
ChrisChrisLoLo: Do we want to create a CSS variable for `425px` in `global.css`, like how we have variables for our CSS colours? Same question applies for the other sizes

src/app/pages/Languages.svelte
```diff
@@ -61,9 +62,8 @@
   }
 
   .languages__container {
-    margin-left: 75px;
```
ChrisChrisLoLo, line 64: Nitpick: What are your thoughts on adding more margin to the left?

I attached a comparison of the two desktop versions
https://imgur.com/a/aQmeVwd

src/app/pages/LandingPage.svelte
```diff
@@ -92,9 +92,9 @@
 
   .masthead {
     /* These hacky variables are here */
-    --wave-height: 222px;
+    --wave-height: 333px;
     --masthead-space-top: 1rem;
-    --masthead-content-height: 13rem;
+    --masthead-content-height: 20rem;
```
ChrisChrisLoLo, line 97: Out of curiosity, I assume that the images got too crowded on mobile with the previous height?

src/app/components/LanguageSources.svelte
```diff
@@ -72,54 +63,58 @@
...
-    {/each}
-  </table>
+      {#each sources as source}
+        <tr>
+          <td>{source.name}</td>
+          <td>{source.size}</td>
+          <td>{englishNameOf[source.type]}</td>
+          <td class="table__row--actions actions">
+            <button class="actions__action btn--inline" on:click={handleEdit}>
+              <img src="/icons/edit.svg" alt="edit" />
```
ChrisChrisLoLo, line 108: These buttons look great!

ChrisChrisLoLo (commented):
I used chrome's device toolbar to see how the site feels on mobile. Knocked it out of the park IMO!

I left some comments to look at. The one I am really interested in is bringing the `max-width` values out as a variable. Feel free to discuss if you'd like to keep it as is.

After that, I'll approve, though you may want to consult @eddieantonio or @zprobs if you'd like a more thorough CSS review.

src/app/pages/LandingPage.svelte
yehee: This one is weird because if you try smaller values it creates a gap in the background:
Screen Shot 2021-02-26 at 6 23 43 AM

`20rem` was just the right value to fill the gap, but let me know if you have a suggestion!

src/app/components/Button.svelte
yehee: Ohh I thought this was such a great idea, but it seems like `var()` doesn't work on media queries ([related StackOverflow thread](https://stackoverflow.com/questions/40722882/css-native-variables-not-working-in-media-queries)). Should we create a .ts file to store constants?

ChrisChrisLoLo: Oh, that's too bad :(
The link you provided was very insightful.

I think a .ts file could definitely work for our purposes! The industry standard though would be to use a CSS preprocessor like PostCSS or SCSS, which I _think_ compiles CSS for us, which can circumvent the problems we are having currently. The counterargument might be that either of these is overkill currently. What are your thoughts on this?

I'm totally fine if we want to mark this a resolved, and then create an investigation task about the best way to go about these types of cases (and talk about pros and cons about `.ts` files vs `.scss` files)

src/app/pages/LandingPage.svelte
ChrisChrisLoLo: Ah, makes sense
I played around with the CSS variables a bit, but I couldn't immediately come up with a viable alternative.
This works!

ChrisChrisLoLo (approved):
LGTM! 

The only action item we might want to consider is opening a new discussion task about how to go about these CSS variables. If you want to make a CSS variable `.ts` file additionally for the time being, feel free to do so

src/app/components/Button.svelte
yehee: Will do! Thanks for looking into it, I'll open up a separate ticket to investigate further into this.

PR #259, opened 2021-02-26
Create cache.yml

### LINKED ISSUE #184: Cache Keyman API on a daily basis

yehee:
## Changes
Created a new GitHub Actions workflow `cache.yml` to fetch from Keyman API and save the output to `src/app/output.json` on a daily basis

Resolves #184


yehee:
@eddieantonio This workflow requires a [GitHub token](https://docs.github.com/en/actions/reference/encrypted-secrets), but we don't seem to have access to the Settings tab to complete the process. Would you be able to take a look?

.github/workflows/cache.yml
```diff
@@ -0,0 +1,36 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Fetch from Keyman API
+        run: |
+          curl https://api.keyman.com/search?q= > src/app/output.json
```
eddieantonio: Let's use a more descriptive name than `output.json` 😂  😂  😂 

```suggestion
          curl https://api.keyman.com/search?q= > src/gen/cached-keyman-api.json
```

Why the `gen` folder? I'm using the convetion of [Android](https://developer.android.com/studio/projects#ProjectView) to place generated files in `gen` instead of with the other source code.

.github/workflows/cache.yml
```diff
@@ -0,0 +1,36 @@
...
+
+      - name: Commit JSON output
+        id: commit
+        run: |
+          git config --local user.email "action@github.com"
+          git config --local user.name "github-actions"
+          git add --all
+          if [-z "$(git status --porcelain)"]; then
```
eddieantonio: `[-z` is invalid syntax because bash is a pile of hacks and `[` is the name of a PROGRAM (which is located at `/usr/bin/[`).  You need a space between the `[` (the program name and `-z`, its first argument.

```suggestion
          if [ -z "$(git status --porcelain)"]; then
```

Note that `[[` is bash-only syntax that makes some `if` statements in bash slightly easier. See: https://jvns.ca/blog/2017/03/26/bash-quirks/#if-statements:~:text=if%20statements,-If

See: https://superuser.com/a/334576/711047

.github/workflows/cache.yml
```diff
@@ -0,0 +1,36 @@
...
+            git commit -m "Update Keyman API" -a
+            echo "::set-output name=push::true"
+          fi
+        shell: bash
+      
+      - name: Push changes
+        if: steps.commit.outputs.push == 'true'
+        uses: ad-m/github-push-action@master
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
```
eddieantonio: It might be easier to use [this action](https://github.com/stefanzweifel/git-auto-commit-action) instead of the current two build steps.

eddieantonio (changes requested):
Good work! What steps have you taken to verify that this GitHub Action will work? CI scripts are notorious for requiring multiple commits and pushes to the default branch to finally get working! 😅

eddieantonio:
> @eddieantonio This workflow requires a [GitHub token](https://docs.github.com/en/actions/reference/encrypted-secrets), but we don't seem to have access to the Settings tab to complete the process. Would you be able to take a look?

`${{ secrets.GITHUB_TOKEN }}` is automatically created, and in fact, does not appear in "Settings » Secrets"!

Please see: https://docs.github.com/en/actions/reference/authentication-in-a-workflow#about-the-github_token-secret

yehee:
> ${{ secrets.GITHUB_TOKEN }} is automatically created, and in fact, does not appear in "Settings » Secrets"!

That's great to know! I'll take a close look at the docs. I've tested locally using [act](https://github.com/nektos/act), although it seems like I would need to set up a few things to make use of GitHub token locally to make sure that the workflow runs smoothly. Would it be possible to get the token just so that I can test it locally?

yehee:
@eddieantonio I made a new commit addressing the feedback! Below is the output from `act` now, seems like I'm seeing the issue mentioned [here](https://github.com/softrams/bulwark/pull/293#issuecomment-695472698) that is yet to be addressed by the dev team. It looks like there might be a few workarounds to make it run locally, but seems like it shouldn't be an issue on the remote repo.
```bash
[Cache Keyman API/build] 🚀  Start image=catthehacker/ubuntu:act-latest
[Cache Keyman API/build]   🐳  docker run image=catthehacker/ubuntu:act-latest entrypoint=["/usr/bin/tail" "-f" "/dev/null"] cmd=[]
[Cache Keyman API/build]   🐳  docker cp src=/Users/alice/code/predictive-text-studio/. dst=/github/workspace
[Cache Keyman API/build] ⭐  Run actions/checkout@v2
[Cache Keyman API/build]   ✅  Success - actions/checkout@v2
[Cache Keyman API/build] ⭐  Run Fetch from Keyman API
|   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
|                                  Dload  Upload   Total   Spent    Left  Speed
100 3799k    0 3799k    0     0   149k      0 --:--:--  0:00:25 --:--:--  918k
[Cache Keyman API/build]   ✅  Success - Fetch from Keyman API
[Cache Keyman API/build] ⭐  Run stefanzweifel/git-auto-commit-action@v4
[Cache Keyman API/build]   ☁  git clone 'https://github.com/stefanzweifel/git-auto-commit-action' # ref=v4
[Cache Keyman API/build]   🐳  docker cp src=/Users/alice/.cache/act/stefanzweifel-git-auto-commit-action@v4 dst=/actions/
| Started: bash /actions/stefanzweifel-git-auto-commit-action@v4/entrypoint.sh
| INPUT_REPOSITORY value: .
[Cache Keyman API/build]   ⚙  ::set-output:: changes_detected=true
| INPUT_BRANCH value:
| The authenticity of host 'github.com (140.82.113.4)' can't be established.
| RSA key fingerprint is.
```

Essentially it hangs because it's missing `known_hosts` directory in the Docker image that they create.

eddieantonio (approved):
Ah, this is much improved! Thanks for testing it in `act`!

PR #262, opened 2021-03-01
Feat/use keyman cache

### LINKED ISSUE #261: Use the cached Keyman API instead of using the Keyman API directly
Follow up to #184

eddieantonio:
Follow-up to #184.

 - I moved @yehee's commits to https://github.com/eddieantonio/cache.predictivetext.studio, hence I deleted the related files here. The reason is that I forgot that pushing to the production is not allowed in this repository, so I made a separate repo _just_ to push to production! Also, Vercel lets me give it the subdomain `cache.predictivetext.studio` which I just think is neat :D 
 - the API now fetches from https://cache.predictivetext.studio/cached-keyman-api.json
 - I also renamed the method because `getDataFromStorage()` was WAAAAAY too generic!

Resolves #261


.github/workflows/cache.yml
```diff
@@ -1,20 +0,0 @@
...
-    steps:
-      - uses: actions/checkout@v2
-
-      - name: Fetch from Keyman API
-        run: |
-          curl https://api.keyman.com/search?q= > src/gen/cached-keyman-api.json
-
-      - uses: stefanzweifel/git-auto-commit-action@v4
-        with:
-          commit_message: Update Keyman API
```
eddieantonio, line 20: This file now lives here: https://github.com/eddieantonio/cache.predictivetext.studio/blob/f19efcf5b50fc90d55c2d9403ad2748046c3100d/.github/workflows/cache.yml

yehee (approved):
Thanks for completing the other part of the task! I was looking into fetching the data from the very same repo, but I couldn't seem to get the right file path to test it. The use of the `cache.predictive.studio` domain makes it clear!

PR #263, opened 2021-03-04
refactor: convert Promise returning function to async

eddieantonio:
Small refactor. Converts Promise-style code to an `async` method.


zprobs (approved):
Big improvement, thanks

PR #268, opened 2021-03-09
Feat/edit and delete dict sources

### LINKED ISSUE #257: Implement editing and deleting uploaded dictionary sources
Currently, we do not have functionality when editing and deleting uploaded dictionary sources in our UI.
Relevant UI
The buttons `Edit` and `Delete` are currently stubbed out.

We need to look at the wireframes created for this feature (#92), resolve any ambiguities about the desired behaviour, and then implement them in the codebase.

Further details can be added below to specify what needs to be edited if the wireframes do not provide enough information.

Depends on #92
Blocks #201

ChrisChrisLoLo:
This pull request allows for Dictionary sources to be updated and deleted.
The PR includes refactorings and tests

Added code:
- Allow for dictionary sources to be edited and deleted in `src/app/components/LanguageSources.svelte`
- Add new method to delete dictionary sources in `src/worker/storage.ts` and in `src/app/logic/delete.ts`

Changed code:
- Remove the `DictionaryEntry` type and all references to it. Converted all previous `DictionaryEntry` objects to type `WordAndCount`. This is because `DictionaryEntry` is a duplicate type of `WordAndCount`.
- Extend `src/app/components/ManualEntry.svelte` so that it can be used to edit existing dictionary sources
- Explicitly set `getLanguageSources` to type `Function`

UI:
Screenshot_20210309_012643

Resolves #257

### DIFF
--- a/cypress/integration/manualEntry.spec.ts
+++ b/cypress/integration/manualEntry.spec.ts
@@ -1,8 +1,52 @@
 describe("Adding sources by direct entry", function () {
-  it("should find a button to press to add source by user entry manually", function () {
+  beforeEach(() => {
     cy.visit("/languages");
     cy.data("languages-sources-btn").contains("Sources").click();
+  });
 
+  it("should create and delete a manual entry source", function () {
+    // create the manual entry
+    cy.data("language-sources-add-sources")
+      .contains("Add Source")
+      .click()
+      .scrollIntoView();
+
+    cy.data("add-sources-splitbtn-direct-entry")
+      .contains("Direct entry")
+      .click()
+      .scrollIntoView();
+    cy.data("manual-entry-input-tablename").type("Common Greetings");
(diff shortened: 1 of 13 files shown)


zprobs (approved):
Tested manually and is working excellently. One thing to consider would be the behaviour of a file which has all its words deleted. Currently files with 0 words remain in the sources list. An alternative would be to delete files with 0 words, although there could be some benefits to keeping word lists with 0 words such as accidental deletions. I'll leave that up to your discretion.

ChrisChrisLoLo:
Thanks for the fast review!

I'll keep the feature as is, as I've noticed that adding additional behaviour can be finicky. This is because the component is being used under two different contexts. I'm okay if the user needs to manually delete the file under these edge cases. If it ever becomes bothersome though feel free to make an additional task!

PR #273, opened 2021-03-12
fix: better typing and test case for Google Sheets

eddieantonio:
Removes `any` from typing.

I also switched the test case to this spreadsheet: <https://docs.google.com/spreadsheets/d/1gShmGH6BJWyAVl8org7G_Wom8XdOYe59bK7tgtjTtJo> which uses greetings in four (non-English!) languages

### DIFF
--- a/test/test-google-sheets-api.ts
+++ b/test/test-google-sheets-api.ts
@@ -6,9 +6,16 @@ import * as c from "chalk";
 import { google } from "googleapis";
 import { readGoogleSheet } from "@worker/read-wordlist";
 import { pathToFixture } from "./helpers";
+import type { GlobalOptions } from "googleapis-common";
 
 dot.config({ path: path.join(__dirname, "../.env") });
 const error = c.bold.red;
+
+////////////////////////////////// Globals ///////////////////////////////////
+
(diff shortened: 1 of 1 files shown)


test/test-google-sheets-api.ts
```diff
@@ -46,37 +81,62 @@ function authorize(credentials: any, callback: any) {
...
       const sheets = google.sheets({ version: "v4", auth });
-      const spreadsheetId = process.env.GOOGLESHEETS_ID || "";
+      const spreadsheetId = SPREADSHEET_ID;
       sheets.spreadsheets.values.get(
         {
           spreadsheetId,
           range: "A1:B",
         },
-        async (err, res: any) => {
+        async (err, res: unknown) => {
```
eddieantonio, line 114: So this kind of sucks, but TypeScript knows the type is something deep within the `googleapis` library, and the types here are, strictly speaking, not that! So it has to be `unknown` and then cast to an appropriate type.

There's probably a better way of doing this, but it involves importing types from deep within `googleapis` which is also not great 🙃

yehee: Absolutely! I definitely see the pain here. I personally don't have a better alternative as of now, perhaps someone could bring in a suggestion when they bump into this later.

yehee (approved):
Thanks for putting these altogether!

PR #275, opened 2021-03-16
Feat/select columns

### LINKED ISSUE #201: Support selecting columns from uploaded spreadsheets
When a spreadsheet (Excel or Google Sheets) is uploaded, it will likely not be in our ideal format, namely, first column for "word" and second column for "count".

Allow the user to select which column from their spreadhseet to use as the word column and which should be the count column.

Predicitve Text Studio should try to make an educated guess by default

ChrisChrisLoLo:
Create a feature that allows users to select the columns they wish to use when uploading a Google sheet or Excel file.
The main changes involved creating a new 'advanced settings' component, as well as extending the functions currently used to upload a spreadsheet.

Major Changes:
- Updated `.env.template` to include the client secret
- Created `UploadAdvancedInput.svelte`, a component where users can input column letters (e.g. A, B, AA) for the 'word' column as well as the 'count' column
- Created `upload-advanced-settings.ts`, which contains functions that can map column letters to zero-based array indices and back
- Create the `UploadSettings` type, which is used to store settings related to the upload (which is currently the index of the columns to use)
- Update existing functions by now passing in a parameter of type `UploadSettings`. These functions can now work with custom columns
- Created appropriate tests

Resolves #201


src/app/components/GoogleSheetsInput.svelte
```diff
@@ -54,14 +61,23 @@
...
       } = await gapi.client.sheets.spreadsheets.values.get({
         spreadsheetId: spreadsheetId,
-        range: "A1:B",
+        range: spreadsheetRange,
       });
-      worker.readGoogleSheet(spreadsheetId, values);
+      const settings = {
+        wordColInd: wordColumnInd,
+        countColInd: countColumnInd,
+      };
```
yehee: Are there any reasons to name these differently? If we make it consistent, we could do
```suggestion
      const settings = {
        wordColumnInd,
        countColumnInd,
      };
```

src/app/components/GoogleSheetsInput.svelte
```diff
@@ -107,14 +123,15 @@
...
+<UploadAdvancedInput bind:wordColumnInd bind:countColumnInd />
 <div class="google-sheets" data-cy="google-sheets-input">
   {#if error}
     <p class:error>{error}</p>
   {/if}
   <InputField
     label="Google Sheets URL"
     id="googleSheetsURL"
-    bind:value={googleSheetsURL}
+    bind:inputValue={googleSheetsURL}
```
yehee, line 131: Just curious, what would be the difference between having `value` and `inputValue`?

yehee (commented):
Great work! One thing I was able to test locally is that when the user passes in the valid index but that column happens to be empty, we still accept that although that would potentially impact the final KMP deliverable. I wasn't sure if that's something that we need to be mindful of.

I also tested TSV files to see some odd behaviors and then realized it only applies to Excel files haha. Just a small comment, but do you think we should make it clear so that it only applies to Excel files?

src/app/components/GoogleSheetsInput.svelte
ChrisChrisLoLo: Good question! `value` isn't defined as a prop in `<InputField>`, but `inputValue` is. https://github.com/eddieantonio/predictive-text-studio/blob/a7203d9a93c84b04cf1015531ac6d66d4821699c/src/app/components/InputField.svelte#L9-L16.

I'm not entirely sure why it worked previously :thinking: . If you have any possible explanation I'd be curious in learning

src/app/components/GoogleSheetsInput.svelte
ChrisChrisLoLo: Fair point :+1:. My reasoning is somewhat subjective. I wanted the `UploadSettings` property to be `wordColInd` rather than `wordColumnInd`, so that way it's more clear to read and type when using settings.`wordColInd`

The reason why I named it `worldColumnInd` in the component prop is that I wanted to be more verbose, as it might be harder for people working with the UI exclusively to know what `col` refers to.

Do you have a preference for one type of spelling? If you think `col` is fairly unambiguous or `wordColumnInd` isn't that much muddier to read and write I can make it consistent either way. It'd be good to hear your opinion on variable naming, so I can be more well versed in rationales for namings.

ChrisChrisLoLo:
Thanks for the feedback! You bring up great points about the UI of the feature

The reason why the app allowed for empty column fields was that it automatically used default columns `A` and `B` as a guess. As you pointed out, this wasn't clearly communicated to the end-user. I've now added UI that indicates what will happen if the user has left the settings fields empty. I also created that file format disclaimer!

Screenshot_20210318_131851

src/app/components/GoogleSheetsInput.svelte
yehee: Agreed! Definitely something to consider, I really don't have a preference on one over the other I personally think `wordColInd` is intuitive enough!

src/app/components/GoogleSheetsInput.svelte
yehee: 😂  My understanding is that the way we're grabbing the URL was by using the `id` prop of the input i.e. `document.getElementById("input-googleSheetsURL")` with `value` never really bound to the input.

src/app/components/GoogleSheetsInput.svelte
ChrisChrisLoLo: @yehee Make sense! I've updated the namings from `.*Column.*` to `.*Col.*` and removed the DOM method in `GoogleSheetsInput.svelte`. I ran uploading via Excel and via Google Sheets to confirm that there were no regressions.

Please let me know if this PR is ready to be merged!

yehee (approved):
Sorry, I thought I've already approved it 😞  Thanks a lot for going through it again!

ChrisChrisLoLo:
No problem! Thanks for approving! :smile:

Issue #187, opened 2021-01-22
Export project (download)


eddieantonio:
We are currently storing data in the browser, using IndexedDB. This works, but there's the persistent threat that [the IndexedDB will be deleted either by the user or by automatically by the browser](https://stackoverflow.com/a/32322186/6626414).

To prevent dataloss, we should have an option to packaged up all the project data and allow the user to download it all locally.

Related to #186

Issue #188, opened 2021-01-22
Mulitple project support


eddieantonio:
Enable the user to be able to work on multiple different languages/models at the same time.

We used to have UI for a sidebar with different languages that resemebled Slack's workspace picking sidebar (see [Figma](https://www.figma.com/file/0ifTqGA2I3Y216eCJxh1rr/Keyman-Predictive-Text?node-id=0%3A1):

![Screen Shot 2021-01-22 at 11 50 38 AM](https://user-images.githubusercontent.com/2294397/105532544-34800b80-5ca8-11eb-8c01-0d9437712b0f.png)

To close this issue, users should be able to create a new project, and switch between their language projects freely.

**NOTE**: this task might be too large for one issue; sub-issues may be required!

yehee:
@eddieantonio I might have missed it, but are there any associations currently in place between projects and files storing wordlist sources? I couldn't seem to find any indication as to which project each file belongs to.

eddieantonio:
@yehee since there only is one "project" right now, there is nothing really explict for multiple projects; however, there is a constant right now that can be used as a "project ID":

https://github.com/eddieantonio/predictive-text-studio/blob/d6486eff9f23041c03d6941e0ca43e037b629fec/src/worker/storage.ts#L10-L13

I believe the files table will need to be changed such that it has a non-null reference to the project it belongs to.

In relational database terms, we'll have two tables:

 * files(**id**, project, name, wordlist)
 * projectData(**id**, langName, bcp47Tag, authorName, modelID, copyright, dictionaryName, version)

And they are related in that there is a many-to-one relationship between files.project and projectData.id. files.project is a foreign key!

There will need to be a data migration that, once files.project is added to the schema, all of the existing files need to have their project set to 0 — the initial project ID.

I don't know if that made any sense >.< please let me know if I can clarify any further!

PR #276, opened 2021-03-17
Feat/download project

### LINKED ISSUE #186: Import project (upload to Predictive Text Studio)
If one has an exported Predictive Text Studio project (see #187), they should be able to upload it again!

To close this issue, there must be UI that allows the user to upload a project from their own computer, and it should appear in Predictive Text Studio, and the user should be able to continue modifying their project and create new models!

zprobs:
Allow users to export their current project as a json file and import json project files to populate their IndexedDB.

Resolves #186 and #187


zprobs:
Unit tests are failing from dexie-export-import. I suppose that is another reason to ditch that library.

src/worker/storage.ts
```diff
@@ -245,6 +246,14 @@ export default class Storage {
     }
     return kmpFile.package;
   }
+
+  async exportProjectData(): Promise<Blob> {
+    return await exportDB(this.db);
+  }
+
+  async importProjectData(data: Blob): Promise<void> {
+    await importInto(this.db, data, { overwriteValues: true });
```
eddieantonio: So these methods export the ENTIRE database — not just the current project. Since the entire DB != one language project, this is doing too much.

eddieantonio (dismissed):
Apart from breaking the unit tests, this PR should add new tests that:

 * export the project
 * clear the database
 * import the project
 * assert that the values have been restored

Please note that the entire database is intended to contain mulitple projects (#188), so exporting the entire database to export one project is inappropriate.

src/worker/storage.ts
zprobs: Ok got it, I'll make these changes

src/app/components/LanguageInfo.svelte
```diff
@@ -18,7 +18,7 @@
   // prevent triggering compilation twice
   let listenForMetaDataChanges: boolean = false;
 
-  onMount(async () => {
+  export async function getMetadata() {
```
eddieantonio, line 23: This is an unusual function to export, especially since it has **side-effects** on the component — namely, if the storedProjectData has changed, then the component will re-render, and also start listening for metadata changes.

It seems to violate the principle of [Command-query separation](https://en.wikipedia.org/wiki/Command%E2%80%93query_separation).

What is the purpose of exporting this function? Can the fundamental functionality needed by callers be handled by a different function? Can there be an Extract Function refactoring that can be used here to isolate the component from side-effects?

src/app/pages/Languages.svelte
```diff
@@ -51,24 +58,60 @@
...
+   */
+  const createAnchor = (href: string, download: string): void => {
+    const anchor = document.createElement("a");
+    anchor.href = href;
+    anchor.target = "_blank";
+    anchor.download = download;
+    // Auto click on a element, trigger the file download
+    anchor.click();
+    anchor.remove();
+  };
```
eddieantonio, line 77: Why use this instead of having this in the markup?

```jsx
<a class="button" download={`${dictionaryName}.kmp`} href={$currentDownloadURL}>
```

src/app/pages/Languages.svelte
```diff
@@ -51,24 +58,60 @@
...
-    anchor.href = $currentDownloadURL;
-    anchor.target = "_blank";
-    anchor.download = `${dictionaryName}.kmp`;
+  /**
+   * Handles downloading the files and metadata stored in IndexedDB
+   */
+  const handleExport = async (): Promise<void> => {
+    const dataString = await worker.exportProjectData();
+    const file = new Blob([dataString], { type: "application/json" });
+    const url = URL.createObjectURL(file);
```
eddieantonio, line 94: 👍🏼

src/common/predictive-text-studio-worker.ts
```diff
@@ -105,4 +105,14 @@ export interface PredictiveTextStudioWorker {
...
+
+  /**
+   * Generates a JSON string of the current IndexedDB
+   */
+  exportProjectData(): Promise<string>;
+
+  /**
+   * Imports data from a JSON string generated using exportProjectData()
+   */
+  importProjectData(data: string): Promise<void>;
```
eddieantonio, line 134: This are good!

src/app/components/LanguageInfo.svelte
zprobs: The reason for exporting is because savedProjectData changes after importing  and so the languageInfo component will need to `fetchAllCurrentProjectMetadata()`. It's called once on mount and then needs to be triggered again by the parent component after a new project is imported, the export is so that it can be called again from Languages.

zprobs: I see what you're saying about the `listenForMetaDataChanges`. I will now set it as false when `getMetadata()` is called so that compilation is not triggered multiple times while setting the languageInfo to the stored/imported value.

src/app/pages/Languages.svelte
zprobs: Yes that's true, changing now

zprobs: The thing with doing the markup is that dictionaryName will need to have a listener to ensure it is up to date. This means sending the info up from LanguageInfo or keeping it in the store. Alternatively doing the anchor programmatically allows us to just retrieve the name when needed.  The createAnchor logic is already used by handleExport too.

eddieantonio: Perhaps a better solution is to refactor: **extract a component** called `<Download href={...}></Download>` that is contains the `createAnchor` logic, but _not_ the logic to create a download name. I feel that creating DOM elements that are never mounted to the document just to use its side-effects is hacky, and may inadvertantly lead to accessibility issues.

src/app/components/LanguageInfo.svelte
eddieantonio: I should have been more explicit: It's _very poor practice_ to name a function "getX" that, in addition to getting X, also does side-effects such as triggering a re-render.

Svelte's reactive assignments ensure that any assignments to component-scope variables, such as `authorName`, `copyright`, etc. will trigger a DOM update: https://svelte.dev/tutorial/reactive-assignments
 

I'm also _very_ concerned about this line:

> It's called once on mount and then **needs to be triggered again** by the parent component after a new project is imported, the export is so that it can be called again from Languages.

A contract like "the parent needs to trigger the re-rendering again of the child is error-prone; even if it were _absolutely required_ to do it this way, such behaviour must be documented. That said, I think an easier soultion is for both elements to subscribe to stores instead, and let Svelte decide when re-rendering should take place.

eddieantonio (approved):
Honestly, the coding practices are not quite up-to standards, however, this PR has been open for far too long. We'll incur a bit of technical debt so that we can continue building upon this work in the meantime.

PR #278, opened 2021-03-18
Add support to modify data read from the uploaded file

### LINKED ISSUE #38: Add support to modify data read from the uploaded file
i.e overwrite count for some words, delete or add new word counts

Specific Requirements: 

* Users can click on the 'edit' icon for any source (Excel, Google Sheet, Manual Entry) to display a table with current values from storage

* Each word will have a row where the word and the count are visible in input text boxes.

* Users can edit values by selecting a text box and tying in a value to update.

  * Values for words cannot be empty and  have to be strings

  * Values for word counts have to be > 0 or empty

  * When users click on an input field, the row should change to match the style for selected input fields.

* Users can click on an 'X' arrow to delete a word from storage.

* There will be a 'new row' at the top of the table where users can type in a new word.

  * When users enter in a value for the row, another new row should appear on top (i.e. there is no 'add word' button and this is done automatically)

* The table name can be edited via a text box.

* The table automatically saves periodically or can be manually saved.

* The table can be closed via a 'close' button

yehee:
It seems like a lot of the features was taken care of already by #257. The list of features included in this PR:
- [x] The table can be closed via a 'close' button
- [x] When users click on an input field, the row should change to match the style for selected input fields.
- [x] When users enter in a value for the row, another new row should appear on top (i.e. there is no 'add word' button and this is done automatically)
- [x] The table name can be edited via a text box.

Resolves #38


src/app/components/ManualEntry.svelte
```diff
@@ -34,33 +34,47 @@
...
+    wordlist = wordlist;
   };
 
-  const addNewRow = (): void => {
+  const addRow = (): void => {
     const newEntry: WordAndCount = ["", 0];
-    rowDataFromManualEntry.push(newEntry);
-    rowDataFromManualEntry = rowDataFromManualEntry;
+    wordlist.push(newEntry);
+    wordlist = wordlist;
```
ChrisChrisLoLo, line 51: What are your thoughts on writing the update like this?:
```suggestion
    wordlist = [...wordlist, newEntry];
```

src/app/components/ManualEntry.svelte
```diff
@@ -34,33 +34,47 @@
...
       return validInput(wordAndCount[WordAndCountInd.WORD]);
     });
     return validTitle && validRows;
   };
 
   const deleteRow = (i: number): void => {
-    rowDataFromManualEntry.splice(i, 1);
-    rowDataFromManualEntry = rowDataFromManualEntry;
+    wordlist.splice(i, 1);
+    wordlist = wordlist;
```
ChrisChrisLoLo, line 45: We could make this:
```suggestion
    wordlist = [...wordlist.splice(i, 1)];
```

src/app/components/ManualEntry.svelte
```diff
@@ -172,13 +190,13 @@
...
-
-    {#each rowDataFromManualEntry as row, i (i)}
+    {#each wordlist as row, i (i)}
       <tr>
         <td>
           <input
             type="text"
-            bind:value={row[WordAndCountInd.WORD]}
+            on:input={addRowIfAbsent(i)}
+            value={row[WordAndCountInd.WORD]}
```
ChrisChrisLoLo: It might be more clear that `wordlist[i][WordAndCountInd.WORD]` is being updated on input if we either:
- rename `addRowIfAbsent()` to `handleWordInput()` or something of the sort
- use `bind:value={wordlist[i][WordAndCountInd.WORD]}` or something similar to indicate that it's this value being changed

What are your thoughts? This is more of a slight nitpick

src/app/components/Upload.svelte
```diff
@@ -20,6 +20,8 @@
 
   /**
    * Re-calculate word count
+   *
+   * Note: A default behavior set to "no action" since no re-calculation is required on landing page.
```
ChrisChrisLoLo, line 29: Good comment!

src/app/components/LanguageSources.svelte
```diff
@@ -38,27 +38,38 @@
...
+    text-align: center;
   }
 
   tr {
     border-bottom: 1pt solid var(--lite-white);
   }
 
+  tr.selected,
+  tr:hover,
+  tr:active {
```
ChrisChrisLoLo, line 67: Nice use of `:hover`, makes the UX feel good!

src/common/types.ts
```diff
@@ -37,6 +37,8 @@ export type DictionarySourceType =
  * A word list stored in the database.
  */
 export interface WordListSource {
+  /** Autogenerated primary key */
+  id?: number;
```
ChrisChrisLoLo, line 41: Could we have a new interface that extends WordListSource?
```ts
interface IndexedWordListSource extends WordListSource {
  /** Autogenerated primary key */
  id: number;
}
```
This way we can know for certain if an object has the `id` property or not. IIRC TypeScript is also able to pass in `IndexedWordListSource` objects into functions that have `WordListSource` parameters without needing any explicit casting

ChrisChrisLoLo (commented):
Good stuff! :grin:

I tried the app out locally: the UX was intuitive and no regressions were apparent. It's great that you made the title editing seamless! My only nitpick in the UX department was that there isn't user feedback after saving an edit. Do you think it's worthwhile to close the table after saving to indicate that the save was successful?

Feel free to take a look at my comments and discuss if you think differently!

src/app/components/ManualEntry.svelte
yehee: Oh yes, I actually tried doing it that way and it didn't work! This would update `wordlist`, but not `tableData.wordlist` if I log their values. I looked into their documentation and it is in fact one of the solutions they suggest: https://svelte.dev/tutorial/updating-arrays-and-objects. But still can't seem to figure out why this approach doesn't work whereas the current approach oddly works. Do you happen to get it to work?

src/app/components/ManualEntry.svelte
yehee: Yes! It used to be doing one job, but now that it's also taking care of updating the value it would make sense to rename it as `handleInput`. Thanks!

src/common/types.ts
yehee: Thanks for bringing it up! I haven't thought about this, I looked into the code and couldn't find a use case for this one just because the `updateManualEntryDictionaryToProject` which explicitly pass in `id` would eventually call `saveFile` which takes in `StoredWordList` which has `id` attribute as optional. Though I might be mistaken, could you elaborate more on this idea?

src/app/components/ManualEntry.svelte
ChrisChrisLoLo: This splice operation doesn't work as I thought it would. I'll resolve this for now

src/app/components/ManualEntry.svelte
ChrisChrisLoLo: Good catch, I had to tweak with the code a bit in order to get something like the above snippet to work :+1:.

What I did was remove the `wordlist` state and worked exclusively with `tableData.wordlist`. My guess is that state somehow gets duplicated when we both have `wordlist` and `tableData.wordlist`.

It might be worth considering removing `wordlist` to prevent these kinds of possible duplicate state bugs from cropping up in the future. What are your thoughts?

```suggestion
    tableData.wordlist = [...tableData.wordlist, newEntry];
```

src/common/types.ts
ChrisChrisLoLo: I think I see where you're coming from! The interface I suggested is just a duplicate of `StoredWordList`, so feel free to disregard that suggestion

To elaborate, my concern with adding optional attributes to `WordListSource` is that there are fewer guarantees about what the object contains. Since `WordListSource` is used in a few areas, there might be times where the object doesn't have an `id` property. This might make it harder to reason about whether or not `id` is available at a certain point of code

I might be misreading, but I think your comment is addressing my concern by stating that there are very few times where `id` will be null.

What are your thoughts on removing the optional `id` property on `WordListSource`, and converting all variables that need the `id` property into type `StoredWordList`? For example, what if we convert `sources` in `LanguageSources.svelte` to `StoredWordList[]` and `tableData` in `ManualEntry.svelte` to `StoredWordList`? This way there might be more clarity as to whether or not the object contains the `id` property

ChrisChrisLoLo (approved):
After chewing it over a bit I think the remaining points are worth considering, but might not be worth blocking over. I feel I might be reviewing more rigorously than what is necessary. I'm still wrangling with what an appropriate level of review is for these types of open-source asynchronous PRs.

If you'd like to discuss or implement the remaining feedback, feel free to do so. If you find the remaining points worth considering, we can also create a new ticket :smile:

The feature works well, and the code is in good shape, so I'll mark this as LGTM!
:shipit:

src/app/components/ManualEntry.svelte
yehee: You're right. I'm also thinking someone else would bump into this later and try to do the same in the future, I might just do that!

src/common/types.ts
yehee: Good point! @eddieantonio I was wondering if you have any comments/suggestions on this topic.

eddieantonio: I agree with @ChrisChrisLoLo. Optional properties are icky, haha! And we _know_ that a wordlist fetched from the database has a valid `id` property. In many other areas of the codebase, the `id` property does not matter, and so they can take a `WordListSource` _without_ the `id` property!

public/global.css
```diff
@@ -140,6 +140,14 @@ a.button:focus {
   padding: .625rem 3.75rem;
 }
 
+* {
+  outline: none;
+}
+
+*:focus {
+  outline: none;
+}
```
eddieantonio: Please **never** do this.

```suggestion
```

It's a usability anti-pattern.

**EDIT**: why? Focus indicators are _necessary_ for folks that navigate predominantly with their keyboards. Not everybody prefers to use a mouse/tap things, so tabbing is a much more effective way of naviating the page. Removing the outlines removes the visual feedback of being able to do this, effectively making it impossible to navigate text fields. Try pressing <kbd>Tab</kbd> when selecting a count! Where does the keyboard focus go? What will happen if you press <Kbd>Enter</kbd> or <kbd>Space</kbd> after pressing tab when selecting the count?

Source:

 - https://www.a11yproject.com/posts/2013-01-25-never-remove-css-outlines/
 - https://hacks.mozilla.org/2019/06/indicating-focus-to-improve-accessibility/#content-main:~:text=Never%20remove%20them
 - https://www.washington.edu/accessibility/checklist/focus/#mobile-sidebar:~:text=Avoid%20Overriding%20Browsers%E2%80%99%20Default%20Focus%20Indicator

eddieantonio (changes requested):

src/app/components/ManualEntry.svelte
```diff
@@ -7,17 +8,16 @@
...
   /**
-   * Disables being able to edit the title
-   * We no longer need this conditional if we work with IndexedDB IDs
+   * Re-calculate word count
    */
-  export let isEditingSource: boolean = false;
+  export let getLanguageSources: Function;
 
-  $: rowDataFromManualEntry = tableData.wordlist;
+  $: wordlist = tableData.wordlist;
```
eddieantonio, line 20: It might be best to avoid aliasing `tableData.wordlist` -- in other words, always use `tableData.wordlist`. :/

public/global.css
yehee: Thanks a lot for bringing this up! I'll definitely make note of that 👍

src/common/types.ts
yehee: I've spent some time investigating this issue, and I think we would have to implement a separate workflow/component to deal with adding a new direct entry vs. editing an existing one. The issue is mainly coming from the fact that they're sharing `tableData` which is currently set to `WordListSource` i.e. missing an `id` attr. If we would like to use `IndexedWordListSource` to enforce having an `id` attribute, then those that are missing an `id` would have a filler value as `id` which also needs to be handled on the worker side.

I might be mistaken, if you have any suggestions as to how to introduce `IndexedWordListSource`, I'd really appreciate it!

eddieantonio: An straightforward way of dealing with this is to create an _empty_ direct entry wordlist, store it, and _then_ modify it. This will be entirely transparent in the UI, but this avoids creating a different workflow for a "new" direct entry wordlist and an existing one; there will _always_ be an existing direct entry wordlist; the only difference is that a new, empty wordlist is created automatically.

If this becomes too involved, I have no problems with spinning this off into its own issue that can be dealt separately outside of this PR.

eddieantonio (approved):

src/common/types.ts
yehee: @eddieantonio Thanks for your insight! I've had a similar idea to that as in when the user switches to a direct entry tab, we would create an empty entry (we then would have access to its `id`) to be modified further. I was worried it would either create a new entry every time the `AddSource` re-renders or we would need to check if there's an empty entry created already for this purpose. I'll create a separate issue referencing this thread to investigate further!

PR #279, opened 2021-03-22
Feat/Make localizable (i18n)

### LINKED ISSUE #185: Make localizable (i18n)
We should enable support for the app to be displayed in multiple different languages. This issue is about removing the hardcoded English strings in the app interface and replacing them with placeholders that can be slotted in with localizations.

To close this issue, the app must have the _ability_ to change its language, either by autodetecting the user's configured language (`Accept-Language` header) or through selecting the desired language from a drop-down box.

This issue _does not_ require a full localization to a different language; just the work to make such a localization possible!

yehee:
- Added `svelte-i18n` package
- Created `src/app/lang` folders to hold the list of language dictionaries e.g. `en.js`, `kr.js`

Screen Shot 2021-03-22 at 12 35 21 AM
Screen Shot 2021-03-22 at 12 34 19 AM

Resolves #185


yehee:
@eddieantonio This is still work-in-progress, but I was wondering if you could take a brief look at it to see if it is on the right track if you have time to do so. Thank you!

eddieantonio (commented):
This is amazing work thus far! 🤩 🤩 🤩 

I'm excited to be able to provide a Spanish localization!

eddieantonio:
Is it possible to automatically set the locale based on the [`Accept-Language` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Language)? If it seems difficult, we could make that an issue for a future PR!

yehee:
@eddieantonio I think it's ready to be reviewed! Some things I should note: I noticed that there are a couple of cases where I had to come up with some sort of a workaround due to the difference between how English is structured vs. Korean. I'm not sure how challenging this would get for other languages...

src/app/App.svelte
```diff
@@ -1,11 +1,23 @@
...
   import LandingPage from "./pages/LandingPage.svelte";
   import Languages from "./pages/Languages.svelte";
   import Privacy from "./pages/Privacy.svelte";
   import Team from "./pages/Team.svelte";
   import Help from "./pages/Help.svelte";
 
+  import en from "./lang/en";
+  import ko from "./lang/ko";
+
+  const [initialLocale] = window.navigator.languages;
```
yehee, line 13: This is a workaround for `getClientLocale` available in `svelte-i18n` package as I couldn't import it somehow, might be related to [this thread](https://github.com/kaisermann/svelte-i18n/commit/31b556bc3f77bc5b581541976a82f898a398c01a#diff-3c1e03e5a3ca9dc9415b55abb4e92f4d1d6975660903414c68d59b658143f642).

eddieantonio: Weird. I was gonna say, it would be nice to use `getClientLocale()` but it appears to [not really do much more than this code](5cf5c938b2ae4641a8271e4c99ceea27c741f76a) anyway 🤔 

Acceptable!

eddieantonio (approved):
This is excellent! LGTM!

Issue #280, opened 2021-03-23
Worker exception when uploading a file without defining a language name first


ChrisChrisLoLo:
A worker exception occurs when uploading a file before setting a language name. This is the following error message.
```
link-storage-to-kmp.ts:12 Uncaught (in promise) Error: Could not fetch the package info; Did you update the package BCP-47 tag?
    at link-storage-to-kmp.ts:12
    at tslib.es6.js:100
    at Object.throw (tslib.es6.js:81)
    at s (tslib.es6.js:72)
```
### How to reproduce:
- Go to https://predictivetext.studio/
- To start with a clean slate, clear the IndexedDB database (in the applications tab in the Chrome console) and Empty Cache and Hard Reload
- Go to the console tab
- Do not select a language, simply upload a file
- Note the error in the console. Note that the language has not been set to `Undefined Language`, indicating that there is a regression and the fix for #227 now no longer works

### Effects:
~~This is potentially causing regressions.~~ For example, this potentially is the reason why #227 and #245 are undone. ~~Other regressions may be occuring.~~
This may also be an edge case that #245 has that the integration tests didn't account for

### Follow up questions/comments:
~~I'm not sure why Cypress didn't catch this. There's a test that checks if #245 works (https://github.com/eddieantonio/predictive-text-studio/pull/245/files#diff-9eb320c5c50bc23dac1750a10d414827094e3ada628329cb187a4a152367209aR112), however, the CI hasn't complained. I will investigate more later when I circle back to this.~~ This may be an unhandled edge case that slipped through the testing cracks

PR #281, opened 2021-03-23
Feat/existing project warning

### LINKED ISSUE #270: Indicate if there is an existing project on the landing page
Right now, the landing page is no different whether you're starting for the first time, or if you've already started a project.

In the case that a project has already been started (wordlist uploaded), indicate this on the landing page. And make "customize" always clickable!

ChrisChrisLoLo:
Create a new notification if the user already has a project started. This notification will show up if the language setting has previously been set.

Screenshot_20210323_231141

### Changes
- Refactor the `ProjectMetadata` type to extend `KeyboardMetadata`, as they share similar attributes. Update relating code appropriately.
- Create UI to display if a project currently exists in `LandingPage.svelte`, as well as the logic required to display it
- Create new `PAGE_URL` enum, so that URLs are no longer hardcoded
- Create a new worker and storage function to query if a project exists. Create appropriate unit tests
- Temporarily disable flaky end-to-end test. This test will fail if someone manually runs the steps in production (see #280 for details). Test also fails when to run in CI. This test should be re-added in #282, as I want to make the tests better before I can better understand why the test occasionally fails. It's possible that mocking out the API call and wiping all state on each test will fix this test

### Notes
- End to end testing for this feature will be added in #282, as some aspects of the landing page tests need to be addressed first

Resolves #270


cypress/integration/landingPage.spec.ts
```diff
@@ -16,26 +16,63 @@ describe("Upload from the the landing page", () => {
...
+  // it("should disable to continue if only language is provided", () => {
+  //   cy.data("landing-page-continue-button")
+  //     .scrollIntoView()
+  //     .should("be.visible");
+  //   cy.data("autocomplete-label").type("zh");
+  //   cy.data("landing-page-continue-button").should(
+  //     "have.class",
+  //     "quick-start__submit-button--disabled"
+  //   );
+  // });
```
eddieantonio: Please use [`it.skip`](https://mochajs.org/#inclusive-tests) instead of commenting out tests!

src/app/pages/page-urls.ts
```diff
@@ -0,0 +1,6 @@
+export enum PAGE_URLS {
+  home = "/",
+  languages = "/languages",
+  team = "/team",
+  privacy = "/privacy",
```
eddieantonio, line 5: This is awesome! Thanks for this!

src/common/types.ts
```diff
@@ -83,17 +83,7 @@ export interface KeyboardDataWithTime extends KeyboardMetadata {
...
-   * The name of language. Could be different than what is fetched by an API.
-   */
-  langName: string;
-
-  /**
-   * The valid BCP-47 tag for the language.
-   */
-  bcp47Tag: string;
-
+export interface ProjectMetadata extends KeyboardMetadata {
```
eddieantonio, line 88: I think the reason this used to be like this is because it matched the Keyman APIs, but their exact APIs are not really quite suited for _this_ app. Thanks for cleaning it up!

src/worker/storage.ts
```diff
@@ -59,7 +59,7 @@ export class PredictiveTextStudioDexie extends Dexie {
        * +------------------+
        */
       projectData:
-        "++id, langName, bcp47Tag, authorName, modelID, copyright, dictionaryName, version",
+        "++id, language, bcp47Tag, authorName, modelID, copyright, dictionaryName, version",
```
eddieantonio: This schema changes **requires** a database migration. Please see: https://dexie.org/docs/Tutorial/Design#database-versioning

## What does a migration look like?

You will leave this line as is. It represents the **original** database schema.

```suggestion
        "++id, langName, bcp47Tag, authorName, modelID, copyright, dictionaryName, version",
```

After `this.version(5)` you will have some code something like this:

```typescript
    this.version(6)
      .stores({  projectData:
        "++id, language, bcp47Tag, authorName, modelID, copyright, dictionaryName, version", })
      .upgrade((transaction) => {
        return transaction.table("projectData").toCollection().modify((project) => {
          project.language = project.langName;
          delete project.langName;
        });
      });
```

---

Although the stakes are fairly low for this project, when modifying an app with an existing database, it's best to do things in database migrations to _upgrade_ the database, rather than to change the schema and require the database to be recreated. It's especially crucial here, since the database exists in each users' browsers, and is not locally managed.

src/worker/storage.ts
```diff
@@ -174,6 +174,17 @@ export default class Storage {
...
 
+  /**
+   * Checks if a project currently exists
+   */
+  async doesProjectExist(): Promise<boolean> {
+    const projectData = await this.db.projectData
+      .where(":id")
+      .equals(PACKAGE_ID)
+      .first();
+    return projectData !== undefined;
```
eddieantonio, line 201: Nice implementation!

eddieantonio (changes requested):
Overall, I really like the changes here! However, there are two issues that must be addressed before merging.

cypress/integration/landingPage.spec.ts
ChrisChrisLoLo: Good catch! I have now used `it.skip` in place of comments

src/worker/storage.ts
ChrisChrisLoLo: Thanks! :grin:

src/worker/storage.ts
ChrisChrisLoLo: Thanks for the insightful comment, learned a thing or two from that! 

That's a really nice feature from dexie, I never considered that front-ends could have databases that needed to be migrated :brain:

I have pushed the requested changes :+1:

src/worker/storage.ts
```diff
@@ -98,6 +98,22 @@ export class PredictiveTextStudioDexie extends Dexie {
           });
       });
 
+    /* Version 6: Conver the `langName` property to `language` in the project table */
```
```suggestion
    /* Version 6: Convert the `langName` property to `language` in the project table */
```

eddieantonio (approved):
LGTM!

PR #284, opened 2021-03-30
Adding copyright button

### LINKED ISSUE #271: Add a © button for the copyright input on the "languages" page
Add a button to append the © symbol to the copyright input in the "Languages" page.

zprobs:
Adds a small button adjacent to the Copyright field in the Languages page. Adds a © on click for better user experience. 

Screen Shot 2021-03-30 at 11 50 42 AM

Resolves #271


src/app/components/LanguageInfo.svelte
```diff
@@ -47,6 +48,10 @@
     let { key, value } = event.detail;
     updateMetadata(key, value);
   }
+
+  function addCopyrightSymbol() {
+    copyright += "©";
```
ChrisChrisLoLo: Should we add logic that only prepends the symbol once?

src/app/components/LanguageInfo.svelte
```diff
@@ -118,7 +134,15 @@
...
       subtext="" />
   </div>
-
+  <div class="copyright-field">
+    <Button
+      size="small"
+      color="blue"
+      isOutlined={true}
+      onClick={addCopyrightSymbol}>
+      add ©
```
ChrisChrisLoLo: nit: We might want to make this `add ©` value use i18n. Here's the relevant PR for localization: #279

ChrisChrisLoLo (dismissed):
Good progress!

Could you prepend the © symbol so that the string looks like this? e.g. `© 2021 GitHub, Inc.`

Currently, if I type in `LangFoundation` and click on the button, I get `LangFoundation©` instead of `© LangFoundation`.

If you can do that and add a cypress test for this feature, I think this should be good to go!

src/app/components/LanguageInfo.svelte
```diff
@@ -76,6 +81,14 @@
     height: 200px;
   }
 
+  .copyright-field {
+    display: flex;
+    align-items: flex-end;
+    margin-bottom: 26px;
```
ChrisChrisLoLo: Nit: we might want to prefer using rem over px according to our style guide: https://github.com/eddieantonio/predictive-text-studio/blob/production/docs/conventions.md#css-prefer-rem-to-px-for-positioning

src/app/components/LanguageInfo.svelte
zprobs: Yes, and you are right that © goes before everything else

src/app/components/LanguageInfo.svelte
zprobs: Added

src/app/components/LanguageInfo.svelte
zprobs: Thanks, changed the other layout px to rem aswell

src/app/components/LanguageInfo.svelte
```diff
@@ -119,7 +140,17 @@
...
   </div>
-
+  <div class="copyright-field">
+    {#if copyrightButtonEnabled}
+      <Button
+        size="small"
+        color="blue"
+        isOutlined={true}
+        onClick={addCopyrightSymbol}>
+        {$_('common.add') + ' ©'}
```
eddieantonio, line 150: It's better to translate the _entire_ button text.

```suggestion
        {$_('common.add_copyright')}
```

I will not block the PR since it requires a decent amount of work, but not all languages will have a 1-to-1 translation for the word "add"; and definitely not all languages place the object after the verb as in "Add[verb] ©[object]". Not all languages use spaces either. This is the Google Translate for "Add ©" in Japanese (unverified, I do not know Japanese, I just listen to a lot of City Pop):

> ©を追加

Notice how © goes first, then there's no space between that and  "を" (-o), the object particle, and 追加 a verb for adding (I'm not sure if this is the right verb!)

eddieantonio (approved):
LGTM! One minor note, but it can be fixed during localization.

PR #286, opened 2021-03-30
Feat/change language page route

### LINKED ISSUE #272: Rename "/languages" to "/customize"?
I'm realizing that the "Languages" page doesn't make much sense anymore. It's like the place where we customize the existing project, so maybe "/customize" is a better name. The Svelte files could be renamed too.

ChrisChrisLoLo:
Change the language page route from `/languages` to `/customize`. Makes changes to relevant files.

Changes:
- Add i18n support to the existing project warning.
- Rename relevant data-cy tags and css classes to use the phrasing `customize` instead of `languages`
- Rename `Languages.svelte` to `Customize.svelte`.
- Update appropriate cypress tests.

Resolves #272


src/app/lang/ko.js
```diff
@@ -50,6 +50,9 @@ export default {
     no_custom_word_column:
       "사용자 지정 단어 열이 지정되지 않았습니다! 기본값은",
     no_custom_count_column: "사용자 지정 수 열이 지정되지 않았습니다! 기본값은",
+    existing_project_warning: "이미 프로젝트가 시작되었습니다!",
```
ChrisChrisLoLo: This used Google translate. I assume a potentially bad translation is better than no translation?

ChrisChrisLoLo: After the standup, the current workflow we'll want to have in the future is to only add translations for languages we are confident in, and then use English as a fallback.

I'll leave the Korean translation for now if @yehee gets an opportunity to review this, going forward I'll work with English only

eddieantonio (approved):
LGTM! 🎉

PR #287, opened 2021-03-30
feat(app): i18n tweaks

eddieantonio:
Improves localzation by:

 - implementing `determinePreferredLocale()`, which finds the first matching locale in common with our translations
 - moving translation files to TypeScript — it's already helped find an erroneous duplicate key!
 - accessing all locales from `./langs` instead of importing each locale manually.

### DIFF
--- a/src/app/lang/en.ts
+++ b/src/app/lang/en.ts
@@ -1,6 +1,8 @@
+import type Localization from "./localization";
+
 export default {
   common: {
-    name: "Predictive Text Studio",
+    app_name: "Predictive Text Studio",
     back: "Back",
     help: "Help",
     excel: "Excel",
@@ -174,4 +176,4 @@ export default {
       team_members: "Team Members",
     },
   },
-};
+} as Localization;
--- a/src/app/pages/LandingPage.svelte
+++ b/src/app/pages/LandingPage.svelte
@@ -336,7 +336,7 @@
           Predictive Text Studio
         </h1>
         <p class="masthead__branding">
-          designed for
+          {$_('page.main.designed_for')}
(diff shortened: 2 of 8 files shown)


src/app/lang/en.ts
```diff
@@ -1,6 +1,8 @@
+import type Localization from "./localization";
+
 export default {
   common: {
-    name: "Predictive Text Studio",
+    app_name: "Predictive Text Studio",
```
eddieantonio, line 5: There were two keys here labeled `name`, and I only realized this after changing this to TypeScript!

src/app/pages/LandingPage.svelte
```diff
@@ -336,7 +336,7 @@
           Predictive Text Studio
         </h1>
         <p class="masthead__branding">
-          designed for
+          {$_('page.main.designed_for')}
```
eddieantonio, line 339: This placeholder was missing.

src/app/pages/Languages.svelte
```diff
@@ -167,7 +167,7 @@
       </a>
       <header class="languages__container--header">
         <div>
-          <h1>{$_('common.name')}</h1>
+          <h1>{$_('common.app_name')}</h1>
           <p>designed for
```
ChrisChrisLoLo: Do we want to replace "designed for" with the following?
```suggestion
          <p>{$_('page.main.designed_for')}
```

ChrisChrisLoLo (approved):
Good stuff, LGTM! :shipit:

src/app/pages/Languages.svelte
eddieantonio: Good catch!

PR #288, opened 2021-04-03
refactor: extract LocalizableDate component

eddieantonio:
Removes the need to translate dates, as Svelte i18n can format dates FOR YOU (probably by using the [`Intl` API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl)).

So now dates are automatically localized! ✨ 
 
Korean localization
Spanish localization


ChrisChrisLoLo (approved):
LGTM! :shipit: 

That's a really API that we get for free

PR #290, opened 2021-04-04
change: tweak copyright button

eddieantonio:
Changes:

 - the entire button text is now localizable
 - the button is always visible; however, the button is disabled when invalid
 - you can add the copyright symbol when the string is empty


ChrisChrisLoLo (approved):
LGTM! :shipit:

PR #291, opened 2021-04-04
refactor: turn split button into a composable component

eddieantonio:
Refactors the old `SplitButton` component into a `ButtonBar`, since we are using it for two tab bars in the application. The button contents themselves have been refactored into a component called `SplitButton`, which obviates the use of a bespoke object to define its properties. This also made it easier to localize the buttons' text.

I have also fixed the current tab highlighting!

Possible further polishing:
 - see if `<SplitButton>` can be removed altogether and replace it with the existing `<Button>` component
 - see if [`$$slots`](https://svelte.dev/tutorial/optional-slots) can be used to apply automatic tab highlighting
 - use non-boolean values for variables that control the currently open tab


zprobs (approved):
Code looking good, tested manually and all is working as expected.

PR #292, opened 2021-04-06
Fix/metadata compilation

### LINKED ISSUE #265: Add Tests for Compilation Callbacks when updating project data and files
See `automatic-compilation.ts`

`compileSuccess` should be `false` immediately after setting language, dictionary name, copy-write etc. or when changing file sources. It should then become `true` once compilation of the KMP is complete.

zprobs:
Adds authorName, copyright and dictionaryName to generated KMP files. Adds cypress test for automatic compilation on metadata changes.

Fixes #265


eddieantonio (approved):
LGTM!

PR #293, opened 2021-04-06
Feat/multiple projects support

### LINKED ISSUE #188: Mulitple project support

### LINKED ISSUE #231: Refactor the file handling to re-use business logic
Currently, the file handling is being copied and pasted for tsv files, Excel files, and Google Sheet files. As more business logic (e.g. ignoring headers) and potential file types (e.g. potentially csv) get added, more code will become more duplicated and bug-prone.

We should design and implement a way that shares as much business logic as possible. Using the Template Method pattern should be a good way of implementing this.

yehee:
## Motivation
- Support multiple projects and add a sidebar on Customize page to ease navigation

## Notable changes
- `{force: true}` option for Cypress testing
- Create `Footer` component
- Delete `BCP47Tag` component
- Merge/replace `addManualEntryDictionaryToProject`, `updateManualEntryDictionaryToProject` with `putDirectEntry`
- Add `project: number` parameter as the project ID of interest to related methods
- `LandingPage` logic:
  - Create a new project data as part of `setLanguage`: calls `putProjectData` that updates existing record if valid project ID is provided; creates a new record otherwise.
  - `putProjectData` returns the project ID which was created/modified
  - If user uploads a new file either by uploading from local or importing from Google Sheets, then `setLanguage` is called
    - If no language is selected just yet, use placeholder value i.e. `Unknown Language`
- `CustomisePage` logic:
  - Created a sidebar which allows user to navigate between projects or create a new empty one
  - By default, the first project on the list is selected/displayed when the page loads
  - A placeholder to indicate no project selected (no project in progress)
- Add `dictionaryName` attribute to `RelevantKmpOptions`
- Move `KMPJsonFileLanguage`, `StoredProjectData`, `StoredWordList` to `@common/types`
- `importProjectData`, `exportProjectData` now imports/exports a list of project data (used to be a single record)
- DB schema for `files` now `++id, name, wordlist, size, project` and `KMPFileData` now `++id, package, project` where `project` is the foreign key to `projectData.id`
- Current DB version: `8`
- No longer referencing `PACKAGE_ID`

Resolves #231, #188


yehee:
@eddieantonio There are still work to do for this PR, but I noticed that some of the cypress test scenarios are failing because "the center of this element is hidden from view":
Screen Shot 2021-04-12 at 10 01 50 AM.

I wasn't sure if others are having similar issues, and they suggest to add `{ force: true }` option as a workaround but just wanted to hear what you think!

eddieantonio:
> I wasn't sure if others are having similar issues, and they suggest to add `{ force: true }` option as a workaround but just wanted to hear what you think!

In general, adding a workaround like `{force: true}` is not the greatest thing 😄 , but it's an acceptable case of incurring technical debt. It happens ¯\\\_(ツ)\_/¯ 

This kind of thing _should_ be fixed, but it's probably not a big deal; it may be a symptom of greater problems, but that's unclear to me at this point. As long as there's **a comment explaining why** the workaround was made, and maybe a `// TODO: do not use force` comment added on there too, it's fine!

cypress/integration/goToTeamPage.spec.ts
```diff
@@ -1,13 +1,16 @@
...
     cy.visit("/");
-    cy.data("team-page-link").contains("About the Team").click();
+    cy.data("team-page-link").contains("About the Team").click({ force: true });
     cy.url().should("include", "/team");
   });
 
   it("should see loaded profiles on the team page", function () {
     cy.visit("/team");
-    cy.wait(500);
+    cy.wait(1000);
```
yehee: I had to adjust values for a few of these to make the test passing, but this could possibly be with my network connection. Please let me know if 500 still works in other machines.

ChrisChrisLoLo: I've had similar problems :+1:

My merged PR increased the times to `2000` since even at `1000` it can fail on rare occasion

yehee:
I should be able to start working on this PR next week as soon as I'm done with my finals 😞 . Any feedback or suggestions would be highly appreciated in the meantime!

eddieantonio:
> I should be able to start working on this PR next week as soon as I'm done with my finals 😞 . Any feedback or suggestions would be highly appreciated in the meantime!

@yehee: you are not obligated to complete this PR. I will give feedback in case you want to finish it.

src/app/pages/Customize.svelte
```diff
@@ -125,17 +146,46 @@
...
+  .customize__sidebar {
+    background-color: var(--gray-dark);
+  }
+
+  .customize__sidebar--project {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    width: 2.5rem;
+    height: 2.5rem;
```
eddieantonio: I was wondering why a single letter needed a flex layout for itself, and then I realized that it's to center it. This works, but a flex layout for a single layout is a bit extra. Here's an alternative that uses vanilla CSS 2 properties (1990s CSS), but `rem` units, because that makes it more robost!

```suggestion
    width: 2.5rem;
    height: 2.5rem;
    
    line-height: 2.5rem;
    text-align: center;
    vertical-algim: middle;    
```

src/app/pages/Customize.svelte
```diff
@@ -209,12 +259,25 @@
  -->
 <main>
   <div class="customize">
+    <div class="customize__sidebar">
```
eddieantonio:
```suggestion
    <nav class="customize__sidebar">
```

src/app/pages/Customize.svelte
```diff
@@ -209,12 +259,25 @@
...
+      {#each projects as project}
+        <div class="customize__sidebar--project" class:selected={project.id === id} on:click={() => id = project.id}>
+          {(project.dictionaryName || "?").substring(0, 1)}
+        </div>
+      {/each}
+      <div class="customize__sidebar--project add" on:click={createProjectData}>+</div>
+    </div>
```
eddieantonio:
```suggestion
    </nav>
```

src/app/pages/Customize.svelte
```diff
@@ -209,12 +259,25 @@
  -->
 <main>
   <div class="customize">
+    <div class="customize__sidebar">
+      {#each projects as project}
+        <div class="customize__sidebar--project" class:selected={project.id === id} on:click={() => id = project.id}>
+          {(project.dictionaryName || "?").substring(0, 1)}
```
eddieantonio: **tl;dr**: `.substring(0,1)`will break for many languages.

There's no good solution for this, but `.substring(0,1)` is, unfortunately, not good to get the first "letter" of the dictionary name. You want the first [extended grapheme cluster](http://www.unicode.org/reports/tr29/#Grapheme_Cluster_Boundaries) of the name, and there's no easy way to do that in vanilla JavaScript at the moment (aside from [dragging in a dependency](https://www.npmjs.com/package/grapheme-breaker) just for this one task).

Also, because JavaScript chooses a rather annoying way of representing Unicode, `.substring(0, 1)` might actually return a _broken_ character: an  unpaired low surrogate. You can replicate this by getting the first "character" of the 💩  emoji:

```
> "💩".substring(0, 1)
'�'
```

src/app/pages/Customize.svelte
```diff
@@ -209,12 +259,25 @@
...
+        </div>
+      {/each}
+      <div class="customize__sidebar--project add" on:click={createProjectData}>+</div>
```
eddieantonio: > Does it _go somewhere_? It's a link.
> Does it _do something_? It's a button.
> Does it code those with a `<div>` anyway? It's a [redacted].
>
> <cite>[[Source]](https://twitter.com/brucel/status/1321746147162152960)</cite>

```suggestion
      <button class="customize__sidebar--project add" on:click={createProjectData}>+</button>
```

src/app/pages/Customize.svelte
yehee: I found [this thread](https://stackoverflow.com/questions/46157867/how-to-get-the-nth-unicode-character-from-a-string-in-javascript) on StackOverflow that might be an alternative to `.substring(0,1)`: `[...s][0]` which will give us the first "character" of `s` assuming `s` is not empty.

eddieantonio: Yes, the ol' [`Array.from(s)[0]` trick](http://eddieantonio.ca/unicode-es6/#/5/4) works here to get the first _code point_. Unfortunately for us, the first code point is not necessarily the first _graphem cluster_, which Matthias Bynens points out in [the accepted answer](https://stackoverflow.com/a/46159939/6626414). 

Try this:

```javascript
> s = ("고추장").normalize("NFD")
'고추장'
> console.log(Array.from(s))[0])
```

Haha, again, this is a very tricky problem to solve in general, for all languages! 😅

yehee: Ohh that's very interesting... Thanks a lot I'll update the code accordingly!

yehee: Is there any reason we're going with `NFD` option for unicode normalization?

eddieantonio: Haha, I would _always_ choose **NFC**. Unless you think "ᄀ" is the correct first letter for "고추장" (I don't know Korean, I just like spicy food).

yehee: That makes sense! I was wondering about that :) Will make the change accordingly!

PR #299, opened 2021-04-08
Feat/upgrade testing

### LINKED ISSUE #295: Clear database on Cypress tests

### LINKED ISSUE #297: Add a placeholder project for Cypress tests
Cypress tests should have a placeholder project that can be used when testing, e.g., the customize page.

@ChrisChrisLoLo 👀 👀 👀

ChrisChrisLoLo:
Improve the existing Cypress tests.

Major Changes:
- Mock out the keyman API responses, as well as the github profile responses. This should allow for more consistent (and faster) testing.
- Clear the indexedDB for all relevant tests. This should make the tests now independent of one another
- Create a new Cypress command to generate projects. The reason why the command manually generates the projects is so that it will always use the latest Dexie Schema version. We cannot get the same guarantee if we were to save the IndexedDB as a file. 
- Update tests to use the above features when needed/desired.

Resolves #297, Resolves #295


cypress/integration/change-metadata.spec.ts
```diff
@@ -1,11 +1,18 @@
 describe("Changing metadata in the language info page", function () {
+  beforeEach(() => {
+    cy.clearLocalData();
+    cy.generateProject();
+  });
+
   it.skip("should find a button to press to add source by uploading file", function () {
-    const languageName = "Makah";
+    const languageName = "’Are’are";
```
eddieantonio, line 8: Should this test case still be skipped?

cypress/integration/change-metadata.spec.ts
```diff
@@ -1,11 +1,18 @@
...
     const authorName = "Eddie";
     const copyright = "© 2018 My Cool Organization";
 
     cy.visit("/customize");
 
+    // Wait for the languages to load
+    cy.wait(1000);
```
eddieantonio: Perhaps mock the keyboard response for this test?

cypress/integration/landingPage.spec.ts
```diff
@@ -10,7 +10,18 @@ describe("Upload from the the landing page", () => {
...
+    cy.intercept("https://cache.predictivetext.studio/cached-keyman-api.json", {
+      fixture: "response-keyman.json",
+    });
+    cy.clearLocalData();
+    cy.wait(500);
+    // make keyman request
+    cy.visit("/");
+    cy.wait(500);
+    // get requested data to appear in dropdown
+    // TODO: We should have Svelte update the dropdown as soon as the languages are loaded
```
eddieantonio, line 22: 👍🏼

cypress/integration/upload.spec.ts
```diff
@@ -10,42 +10,24 @@ describe("Upload from the the landing page and Download KMP", function () {
...
+    cy.clearLocalData();
     cy.task("clearDownloads");
     cy.allowUnlimitedDownloadsToFolder(downloadFolder);
-    cy.visit("/");
   });
 
   it("should find a button to press to upload a file", function () {
+    // Load languages
+    cy.visit("/");
+    cy.wait(500);
```
eddieantonio, line 21: Are these waits for `cy.intercept()`?

eddieantonio (commented):
In general, this is great! I really like the new commands!

However, I'm confused about all the `cy.wait()` invocations everywhere. Is that for `cy.intercept()`? Is there a way to reduce the number of explict `cy.wait()` calls?

---

Also, GitHub won't let me comment on it, but what's up with `cypress/fixtures/indexedDB.ts`?

ChrisChrisLoLo:
Thanks for the quick feedback!

I have removed the unused ts file.

Some of the waits are for the cy.intercept, yes!
However, some of them try to make fixes to race conditions that are occurring in the test runner. I'm a bit frustrated as these tests are working locally, but seem to have flaky behaviour in the runner (for example, the app doesn't _seem_ to load the indexedDB in time in the UI unless you wait for a second) :thinking:.

You can look at previous test runs that failed to see examples of this race condition:
e.g. https://github.com/eddieantonio/predictive-text-studio/runs/2301300974

The race conditions don't make sense to me, so maybe I need to do something in terms of Promises with the new Cypress commands? I'll investigate more into this. If I can figure out what's causing the race-conditions, I should be able to remove all `wait()`s unrelated to `cy.intercept()`.

cypress/integration/upload.spec.ts
ChrisChrisLoLo: EDIT: More succinct answer in the next comment below

This `wait` is for the race conditions I'm experiencing. While I cannot get a direct error message, it seems that there are moments where the app will not load things in time from the IndexedDB. This behaviour didn't occur before, so I'll see if I need to make the clearing and generation of the IndexedDB blocking or something of the sort

cypress/integration/change-metadata.spec.ts
ChrisChrisLoLo: ~~That could work! I'll need to find a way to implant languages into the Svelte application. If making the IndexedDB generation and deletion blocking doesn't work, I'll investigate if mocking could be another approach to this problem.~~

cypress/integration/change-metadata.spec.ts
ChrisChrisLoLo: Good catch! For some reason this test cannot have it's edits persist. I'm not sure why this is. I'm looking into it

ChrisChrisLoLo:
Poking around the codebase a bit, it seems like we have `cy.wait()`s to allow us to load the project in time.
https://github.com/eddieantonio/predictive-text-studio/blob/76e2f36cb2ecc1408f7a5fa408cff8d211e65f71/cypress/integration/copyright-button.spec.ts#L10

Maybe since the `wait()`s that I added are still necessary?

I'll keep investigating for a small while though. Is there a way you can think of that can allow us to avoid these waits?

cypress/integration/change-metadata.spec.ts
ChrisChrisLoLo: The test seems to fail here, as the changes made here don't persist when the page reloads. I can't explain why this would be the case

ChrisChrisLoLo: ~~This test should work now~~. Test works locally, but not on the CI runner

cypress/integration/change-metadata.spec.ts
ChrisChrisLoLo: I think the keyboard response here is mocked out (the intercept gets set up in `cy.generateProject()`). My guess as to why this `cy.wait()` needs to now be here is that the mocked response doesn't get cached, and therefore Cypress needs time to intercept the request that the Svelte app sends out

cypress/integration/upload.spec.ts
ChrisChrisLoLo: Yes, these waits are for `cy.intercept()`. I'm fairly sure the reason why these `wait()`s are required now is that the keyman keyboards get cleared on every test. Since the keyboard gets cleared, an API request needs to be made and get intercepted, which can take milliseconds to complete.

What I will try later today is to make a command that clears everything _except_ the keyman keyboards. Hopefully, this should remove the need for these waits, while also clearing out the previous projects

ChrisChrisLoLo:
@eddieantonio I have made it so that the Keyman keyboards won't get cleared out. This means that few API requests should now be made, meaning in theory, we don't need to wait anymore.

Despite this change, however, some tests still seem to be flaky, indicating that
1. some of the additional `wait()`s will still be necessary
2. some `wait()`s need to be increased to prevent race conditions.

For example, https://github.com/eddieantonio/predictive-text-studio/runs/2310173495 failed but https://github.com/eddieantonio/predictive-text-studio/runs/2310205425 didn't despite trivial changes.

I'll tinker around a bit, but the current `wait()`s may be required, and some future tuning with the `wait()` times may also be required

cypress/support/commands.js
```diff
@@ -55,3 +57,78 @@ Cypress.Commands.add("readZip", (filename) => {
...
+  localStorage.clear();
+  return cy.window().then((window) => {
+    return window.indexedDB.databases().then((databases) => {
+      for (var i = 0; i < databases.length; i++) {
+        if (databases[i].name === "dictionary_sources") {
+          new Dexie(databases[i].name).open().then(function (db) {
+            db.tables.forEach(function (table) {
+              // console.log("Found table: " + table.name);
+              if (table.name !== "keyboardData") {
+                // console.log(table.name);
```
eddieantonio:
```suggestion
              if (table.name !== "keyboardData") {
```

eddieantonio (approved):
Please remove the commented out `console.log`s, then LGTM!

